Issue #14, opened 2017-08-28
Select bounding box for each peak


pmelchior:
Currently the model of each component covers all pixels in the image. That's overkill for the evaluation of the likelihood gradients. Instead one could work with a subset of pixels, loosely based on previously detected footprints.

One can think of a few ways of doing that:
1) Each component exists only in its box, so that constraints only have to check for that box (we'd have to insist that the models are zero outside of that bounding box otherwise each component can do whatever it pleases outside of its own box). For the all procedures that require the actual model, e.g. likelihood gradients, the components need to be translated accordingly. That means that S is internally always represented by a vector (not a matrix), which we already partially implement in the dot products of block-diagonal constraint matrices, and the generation of the model is more involved. The downside is that we need to build the constraint matrices separately for each component (or work with fixed box size increments to reduce that number). It seems to me that this will increase initialization time but reduce runtime.
2) Use a sparse matrix form for S that allows us to quickly update a set of predefined elements (for every component: those inside of the box, the rest is zero) as long as we don't change the structure of the matrix. The models would automatically have the right format, and the changes to the code would be minimal. There is probably a slightly larger runtime cost. This approach seems fairly straightforward to me, but I can't seem to find a way to do this within `scipy.sparse`.

In both cases, we'd automatically prevent the degeneracy between two remote objects having the same color.

pmelchior:
Here's a different thought that should solve all of the problems we've had about separable objects
run bSDMM with a extended set of variables `[S1,S2,...,SK,A]`, where Si is the spatial part of i-th object. This way they can be treated independently (in whatever boxes we want to have) and the convergence can be checked for any of them separately. This last part should be very good for diagnostic purposes.

This requires a few changes.
1) deblender calls `proxmin.bsdmm` directly.
2) a new class for building the model and calculating likelihood gradients.
3) initialization of constraints for S needs to be split.

2) will be most important. There are several redundant computations for the likelihood gradients. By pulling those together in a class, we can store the results of previous calculations and control when updates to internal variables (such as the fully assembled S matrix, or the A matrix) are done.
It would also be the interface for the user to inspect and render the deblender outputs.

[...]

PR #20, opened 2017-12-11
Scarlet

pmelchior:
This PR implements the per-source box treatment (#14) and brings with it an entirely updated API.

It uses the mathematically correct step sizes, PSF and step size caching, fluent design, and numerous other improvements. It is expected to perform faster for images with many small objects, and slower if the initial box sizes are much too small (so that several box size updates are necessary).

Because this is qualitatively a different animal I suggest that the current `master` branch is preserved under a different name, such as `old_nmf`.


scarlet/blend.py
```diff
@@ -0,0 +1,659 @@
...
+
+    def component_of(self, m, l):
+        """Search for k index of source m, component l
+
+        This is the inverse of source_of, and returns `k` for the given
+        pair `(m,l)`.
+        """
+        for k in range(self.K):
+            if self._source_of[k] == (m,l):
+                return k
```
fred3m: I think this is _slightly_ faster using
```
k = self._source_of.index((m,l))
if k<0:
    raise IndexError
return k
```

scarlet/blend.py
```diff
@@ -0,0 +1,659 @@
...
+            self._edge_flux = np.zeros((self.M, 4, self.B))
+
+        # top, right, bottom, left
+        self._edge_flux[m,0,:] = np.abs(model[:,:,-1,:]).sum(axis=0).mean(axis=1)
+        self._edge_flux[m,1,:] = np.abs(model[:,:,:,-1]).sum(axis=0).mean(axis=1)
+        self._edge_flux[m,2,:] = np.abs(model[:,:,0,:]).sum(axis=0).mean(axis=1)
+        self._edge_flux[m,3,:] = np.abs(model[:,:,:,0]).sum(axis=0).mean(axis=1)
+
+    def resize_sources(self):
+        """Resize frames for sources (if necessary)
```
fred3m: We might want to come up with a better way to estimate how much larger the source needs to grow. That would help in cases where you have a very bright object that needs a much larger footprint.

pmelchior: I agree. I tried to do this with the relative 25% increase. It's not trivial to know how to do the increase because we have no immediate way of knowing how much larger a source could be.

scarlet/source.py
```diff
@@ -0,0 +1,481 @@
...
+        shape: tuple
+            Shape of the frame that contains the source. This can be (and usually is)
+            smaller than the size of the full blend
+        K: int, default='1'
+            Number of components with the same center position
+        psf: array-like or `~scarlet.transformations.GammaOp`, default=`None`
+            2D image of the psf in a single band (Height, Width),
+            or 2D image of the psf in each band (Bands, Height, Width),
+            or `~scarlet.transformations.GammaOp` created from a psf array.
+        constraints: dict, default=`None`
```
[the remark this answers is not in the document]
pmelchior: I suggest, we just pass a list of constraint dicts for the time being. If the list is just one dict, we clone it K times. I'll make that change now.

PR #29, opened 2018-02-06
Trafo cache

pmelchior:
Transformation matrices get on-the-fly cache to reduce construction cost.

As a side consequence, `Blend` needs a new keyword `source_sizes` to know which sizes are allowed according to a predefined list.

To work best, `Source` initialization needs to respect the same size options, which needs to get fixed when #28 is merged. However, this PR can be merged with master without conflicts.


pmelchior:
Let open a new issue for the max box size problem

Issue #30, opened 2018-02-06
Source initialization needs to respect size options


pmelchior:
This is an extension of #18. 

With #29 merged, the size updates are taken from a fixed set of box sizes, which can be configured as argument of `Blend`. However, `Source` is still initialized in some way, having some box size. That defies the purpose of having a limited set that can efficiently be cached.

Proposal: Define global variable `scarlet.source_sizes` with a small helper function to decide which is the best available size. Both is currently part of `Blend`, but it needs to move up to package level.

Bonus points: `Blend` carries several other config parameters that change the behavior of the fitting process, but are unlikely to be altered between consecutive instantiations of `Blend`. I would therefore rather define all of them at the package level. This then cleans up the `Blend` class.

Thus, add the following to `__init__.py`:
```
config = {
  refine_skip=10, 
  source_sizes = [15,25,45,75,115,165],
  edge_flux_thresh=1.,
  exact_lipschitz=False
}
```
and have `Source` and `Blend` load parameters from there.

pmelchior:
This came up during PR #29: There should be some check if the maximum source size is consistent  with the size of the image. If not a fallback option should be added to the helper function.

PR #32, opened 2018-02-07
Source cache

pmelchior:
Solves #30 by introducing a config module, so that source sizes (and other aspects of scarlet) can be shared across the package.

### DIFF
--- a/scarlet/__init__.py
+++ b/scarlet/__init__.py
@@ -3,4 +3,4 @@ from proxmin import nmf as nmf
 from .constraints import *
 from .source import *
 from .blend import Blend
-from . import config
+from .config import Config
--- a/scarlet/__init__.py
+++ b/scarlet/__init__.py
@@ -3,3 +3,4 @@ from proxmin import nmf as nmf
 from .constraints import *
 from .source import *
 from .blend import Blend
+from . import config
--- a/scarlet/blend.py
+++ b/scarlet/blend.py
@@ -645,8 +645,8 @@ class Blend(object):
         for m in range(self.M):
             if not self.sources[m].fix_frame:
                 size = [self.sources[m].Ny, self.sources[m].Nx]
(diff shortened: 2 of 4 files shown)


fred3m:
I like the idea of having a configuration that can be used in multiple places, like `Source` and `Blend` objects, but I'd rather see a config class that can be passed around than a module. The main reason is to make it easier to persist the configuration used. With the scheme in this commit, the configuration used for a particular blend is lost if the same file creates multiple blend objects with different configuration parameters, which is reasonable when running the deblender on a new dataset and trying to find the optimal values for parameters.

pmelchior:
It's a bit of a special case, but I see your point.

Also, it would allow to fix the source_sizes (namely at instantiation) instead of every time any piece of the code needs to access it. Wait for another commit ...

pmelchior:
In retrospect, I should have thought about this a little longer. The idea of passing `config` around is counter to the intended purpose. It would need to be passed around only if there can be multiple instances in parallel, which isn't what will happen here. Therefore, I don't think `config` should be a class because you don't want to (have to) instantiate it, so it's all about static methods etc., and at that point you might just do it as a module, which is what I've done. That's the python way.

Moreover, I don't think the problem you mention ever arises in a production environment where persisting configs would be useful.

fred3m:
All of the configs in the LSST stack are in a ConfigClass, not a module, for the purpose of persisting them. It is also useful for testing, where there may be multiple different configurations used in the same module to test how different parameters affect the result.

pmelchior:
The purpose in the LSST stack probably arises from having multiple concurrent instances of `ConfigClass`. I can't see that being the case for the deblender.

As for testing, I doubt that you'd need to persist.

Anyway, should this ever become an issue, then we'll address it. This PR is about preventing misses in the transformation cache.

PR #33, opened 2018-02-08
Display

fred3m:
This implements a display module with basic functionality that is useful to view deblender results.

It includes the creation of a class `Asinh`, which can be used in matplotlib in the following way:
```
asinh = scarlet.display.Asinh(vmin=0, vmax=10, Q=20)
plt.imshow(img, norm=asinh)
plt.colorbar()
plt.show()
```
which displays the image `img` using an arcsinh stretch, and an appropriately scaled and labeled colorbar.

The module also includes a method to convert an image (or deblender model) into an RGB color image that can be displayed in matplotlib. For example if `blend` is the deblender output:
```
img = blend.get_model()
rgb = scarlet.display.img_to_rgb(img, vmin=0, vmax=np.max(img)/2, Q=50, filter_indices=[3,2,1])
plt.imshow(rgb)
plt.show()
```
will show a color image of the model.

### DIFF
--- a/scarlet/__init__.py
+++ b/scarlet/__init__.py
@@ -3,4 +3,3 @@ from proxmin import nmf as nmf
 from .constraints import *
 from .source import *
 from .blend import Blend
-from . import display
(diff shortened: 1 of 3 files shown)


pmelchior:
Please remove the import from `__init__.py` because this code needs to run on headless machines (or more general: one some without mpl)

PR #39, opened 2018-03-15
Smooth

fred3m:


fred3m:
Note: this issue was closed because the smooth, convex constraints fail to converge in a few cases and there is no pressing need to implement them. If a situation arises where we find it necessary to implement a smooth constraint we can revisit this branch and investigate what is causing the failure.

PR #40, opened 2018-03-19
Symmetry

fred3m:


pmelchior:
Thanks for adding the direct symmetry operator.

However, the `repeat` argument doesn't make sense for individual constraints. It's only relevant for `AlternatingProjections` and cannot predicted by single constraints: it's required only if multiple constraints are hard to enforce simultaneously. Please remove it from the base class and the derived classes.

fred3m:
I had a feeling that you'd push back on this, but there's no good way to implement this. If I didn't include `repeat` in the base class then the API would be clunky, because we'd have to initialize sources like this:

```
import scarlet.constraints as sc
constraints = [sc.PositivityConstraint() & sc.SoftSymmetryConstraint() for n in range(len(catalog))]
for c in constraints:
    c.set_repeat(3)
sources = [scarlet.Source((src["y"], src["x"]), constraints[n]) for n,src in enumerate(catalog)]
```

as opposed to

```
sources = [source.Source((src["y"], src["x"]), constraints = sc.PositivityConstraint(repeat=3) & sc.SoftSymmetryConstraint()) for src in catalog]
```

using the current API. There are two advantages to this:

1. The entire source initialization can only be written on one line using the current API
2. There is no good way to combine constraints if we don't include `repeat` in the base class. You have to combine constraints, which will create an `AlternatingProjections` object with the default (`repeat=1`). Then you have to add an update method to `ConstraintList` which creates a new `AlternatingProjections` object with a new `repeat`.

You'll notice that I only included `repeat` in the inherited classes that have a `prox_morph` or `prox_sed` proximal operator, not the `prox_g` operators, since they will never be `AlternatingProjections` anyway. Then the `ConstraintList` takes the _maximum_ `repeat` value. I think that this is the right way to do this.

PR #41, opened 2018-04-01
Replace sparse matrices in Gamma with analytic operators

fred3m:


scarlet/transformations.py
```diff
@@ -15,147 +16,437 @@ def check_cache(name, key):
...
-        """Build Tx, Ty, P, Gamma
+        self.img = np.array(img)
+        self._coords = np.array(coords)
+        self.slices, self.inv_slices = get_filter_slices(self._coords.reshape(-1, 2))
+    
+    @property
+    def T(self):
+        """Transpose the filter
+        """
+        return LinearFilter(self.img, -self._coords)
```
pmelchior: That's incorrect: The coordinates aren't negated, they are flipped, that is x and y are interchanged.

fred3m: I thought the same thing as you at first, but this isn't matrix coordinates, this is the coordinates of the image that is being applied as an operator. So (0,1) is the pixel to the right of the current pixel (0,0), and (0,-1) is the pixel to the left of the current pixel, which is the transpose of the operator.

I verified this by using the old operators and transposed operators and applying them to the same matrix, making sure that they gave bitwise identical results.

scarlet/transformations.py
```diff
@@ -15,147 +16,437 @@ def check_cache(name, key):
...
+            (so [0,0] is the current pixel).
         """
-        self.psf = psf
-        self.cache = {}
-
-    def _make_matrices(self, shape, offset_int):
-        """Build Tx, Ty, P, Gamma
+        self.img = np.array(img)
+        self._coords = np.array(coords)
+        self.slices, self.inv_slices = get_filter_slices(self._coords.reshape(-1, 2))
```
pmelchior: This will not work if `img` is in fact an image, in which case `coords` are redundant. This might also get potentially inconsistent with `self.img.reshape` in L136.

scarlet/transformations.py
```diff
@@ -15,147 +16,437 @@ def check_cache(name, key):
...
-                                      shape=(size, size))
-        ty_plus = scipy.sparse.diags([1., -1.], offsets=[offset_int[0]*width, (offset_int[0]-1)*width],
-                                     shape=(size, size))
-
-        P = self._adapt_PSF(shape[1:])
-        return tx,tx_plus,tx_minus,ty,ty_plus,ty_minus,P
-
-    def __call__(self, pos, shape, offset_int=None):
-        """Get the operators to translate source
+        return apply_filter(X, self.img.reshape(-1), self.slices, self.inv_slices)
```
pmelchior: I find this entire construct rather indirect: `dot` calls `apply_filter` which then calls yet another `apply_filter`. Do you have reasons for this modularization?

pmelchior: I would simple code the entire actual operation in `dot`

fred3m: That is what I did originally but I made it modular for 2 different reasons.

1. The `apply_filter` function call was put outside the class in case we find a way to optimize it in C++. 
2. The `apply_filter` is a method in the class in case a particular filter wants to override this method. For example, when I was calculating the runtime using bitmasks and alternative versions specific to the translation operators I wanted to override `apply_filter` but not `dot`, because I did not want to have to reimplement the logic in `dot` that decides how to apply the filter to other filters vs a matrix.

So I'd like to leave it as is to make the framework more flexible.

scarlet/transformations.py
pmelchior: Yes, my mistake. This is the transpose of the transformation matrix, not of the filter image.

scarlet/transformations.py
pmelchior: Not quite. There are two options here: 
1) `img` is a 2D filter image, then it need no coordinates, but constant reshaping in `dot()`
2) `img` is a 1D list of values, together with a list of 2D coordinates, in which the coordinates are necessary, but there is no need for reshaping in `dot()`.

I think we should store internally the latter version and make sure that we can parse both.

scarlet/transformations.py
pmelchior: I can see the reasoning behind argument 1, but not for 2. I'd say since these calls are time-critical, I'd streamline them as much as possible. Three function calls to achieve one thing isn't necessary.
Plus: See my comment above about the `reshape(-1)`.

scarlet/transformations.py
```diff
@@ -15,147 +16,437 @@ def check_cache(name, key):
...
+            (or chain of filters) then a new `LinearFilterChain` is
+            returned with this one prepended.
+        """
+        if isinstance(X, LinearFilter):
+            return LinearFilterChain([self, X])
+        elif isinstance(X, LinearFilterChain):
+            X.filters.insert(0,self)
+            return X
+        else:
+            return self.apply_filter(X)
```
pmelchior: Let this actually be the call to the module-level `apply_filter`, which may or may not be implemented in C++.

scarlet/transformations.py
```diff
@@ -76,27 +76,76 @@ class LinearFilter:
...
+                    raise ValueError(msg)
+                center = [self.values.shape[0]//2, self.values.shape[1]//2]
+            self.center = center
+            x = np.arange(self.values.shape[1])
+            y = np.arange(self.values.shape[0])
+            x,y = np.meshgrid(x,y)
+            x -= center[1]
+            y -= center[0]
+            coords = np.dstack([y,x])
+        self._coords = np.array(coords)
```
pmelchior: Thanks for making the adapter for 2D images.

I would prefer if we only had one version of `values` and `coords` stored in the class, namely the flattened ones. You can regenerate the full ones if needed from them anyway, so no reason to keep them around. I would therefore replace the lines from here to end of method with
```
self._flat_values = np.array(values).reshape(-1)
self._flat_coords = self._coords.reshape(-1,2)
assert ...
# remove elements with zero value
non_zero = self._flat_values != 0
self._flat_values = self._flat_values[non_zero]
self._flat_coords = self._flat_coords[non_zero]
self._slices, self._inv_slices = get_filter_slices(self._flat_coords)
```
and run the entire method as `__init__` because there's nothing left in init to do (since `self.values` isn't needed anymore). Note that the transpose operation could work on the flattened values and coordinates just as well.

PR #43, opened 2018-04-09
Implement Eigen in place of numpy filters

fred3m:


fred3m:
Before we merge, I'll remove the "cpp" arguments that allow the user to turn off Eigen. This is just for testing until we are satisfied that the code installs and executes correctly.

scarlet/operators_pybind11.cc
```diff
@@ -21,66 +24,69 @@ void prox_monotonic(
...
-    py::array_t<double> &result
+// Apply a filter to an image
+template <typename M, typename V>
+M apply_filter(
+    Eigen::Ref<const M> image,
+    Eigen::Ref<const V> values,
+    Eigen::Ref<const EigenVector> y_start,
+    Eigen::Ref<const EigenVector> y_end,
+    Eigen::Ref<const EigenVector> x_start,
+    Eigen::Ref<const EigenVector> x_end
```
TallJimbo: It probably doesn't matter, but you might be able to make this go a bit faster by having this take an existing NumPy array (via `Eigen::Ref`) as an output parameter - returning a true `Eigen::Matrix` object will involve a deep copy when converting to NumPy.

setup.py
```diff
@@ -73,6 +75,50 @@ def cpp_flag(compiler):
...
+            import tarfile
+            import requests
+            import tempfile
+
+            print("Downloading and extracting Eigen headers")
+            url = "http://bitbucket.org/eigen/eigen/get/3.3.4.tar.gz"
+            download = requests.get(url)
+            tarball = tarfile.open(mode="r:gz", fileobj=BytesIO(download.content))
+            files = [f for f in tarball.getnames() if f.startswith("eigen-eigen-5a0156e40feb/Eigen")]
+            # Exctract only the header fiels from the archive
```
TallJimbo: typo

setup.py
```diff
@@ -39,7 +39,9 @@ def __str__(self):
         ['scarlet/operators_pybind11.cc'],
         include_dirs=[
             get_pybind_include(),
-            get_pybind_include(user=True)
+            get_pybind_include(user=True),
+            # Path to Eigen submodule
```
TallJimbo: Might be confusing to call this a submodule since it isn't actually a *git* submodule.

setup.py
```diff
@@ -109,6 +155,6 @@ def build_extensions(self):
   keywords = ['astro', 'deblending', 'photometry', 'nmf'],
   ext_modules=ext_modules,
   install_requires=['proxmin>=0.5.0', 'pybind11>=1.7', 'numpy', 'scipy'],
-  cmdclass={'build_ext': BuildExt},
+  cmdclass={'develop': InstallEigen, 'build_ext': BuildExt, 'install': InstallEigen},
```
TallJimbo: I don't know setuptools well enough to know for certain what this is doing, but you shouldn't have to install the Eigen headers after you've built the module.

TallJimbo:
When trying out the build procedure for myself, in addition to the dependencies listed I also needed the `requests` package.  After installing that (I used conda), everything worked.

pmelchior:
I think `setuptools` has an option called `setup_requires`, which should list `requests` to avoid the problem or at least provide a straightforward error message.

scarlet/operators_pybind11.cc
fred3m: Good call, it is slightly faster when using PSF deconvolution now.

setup.py
fred3m: This is just saying if you use `python setup.py develop` or `python setup.py install`, run the `InstallEigen` class. The order doesn't matter, which is why it's a dict.

fred3m:
I didn't realize that `requests` was not part of the standard library. I guess it should be ok to add this as a required dependence, since most astronomy users are not likely to have a system install of `Eigen` and will need to download the headers.

pmelchior:
Yes, I prefer that over shipping Eigen with scarlet

PR #44, opened 2018-04-25
Psf

fred3m:
Implement PSF matching in scarlet


scarlet/source.py
```diff
@@ -597,3 +597,114 @@ def make_initial(self, img, bg_rms, thresh=1., symmetric=True, monotonic=True, c
             # keep the peak sed
             logger.INFO("Using peak SED for source at {0}/{1}".format(self.center_int[0], self.center_int[1]))
         return sed.reshape((1,B)), morph.reshape((1, morph.shape[0], morph.shape[1]))
+
+class PSFDiffKernel(Source):
```
pmelchior: If you derive this from `ExtendedSource`, you can use all of its `make_initial`.

pmelchior: The only thing needed as a SED that is set to `(0,0,1,0,0)` or the like, but you can use the morphology from `ExtendedSource`

scarlet/psf_match.py
```diff
@@ -1,83 +1,168 @@
...
+        `psfs` to `target_psf`, where `diff_kernels` is an array
+        of the `Source.morph` for each source in `psf_blend`.
+    """
+    config = Config(refine_skip=100, source_sizes=np.array([np.max(psfs.shape[1:])]))
+    center = np.array([psfs[0].shape[0] // 2, psfs[0].shape[1] //2], dtype=psfs.dtype)
+    if constraints is None:
+        constraints = sc.SimpleConstraint() & sc.L0Constraint(l0_thresh)
+    if cutoffs is None:
+        cutoffs = [0]*len(psfs)
+    target_psfs = np.array([target_psf]*len(psfs))
```
pmelchior: That is not necessary because `Gamma` in the source will expand this if it's one PSF for all bands.

scarlet/psf_match.py
```diff
@@ -1,83 +1,168 @@
...
+            params.append(np.mean(all_params[:,4])) # amplitude 2
+            params.append(np.min(all_params[:,5])*0.6) # sigma 2
+    else:
+        params = extract_values(all_params)
+    target_psf = func(coords, y0, x0, *params)
+
+    # normalize the target PSF
+    target_psf = target_psf/np.sum(target_psf)
+    return target_psf
+
```
pmelchior: Can you move `PSFDiffKernel` from `sources` to here. It's a special-purpose source, which should best not be used outside of the PSF matching.

scarlet/psf_match.py
fred3m: That was true, but the new Gamma has no notion about the number of bands other than from the length of `psfs`. However, I did add logic in `Source` to use a single PSF in all bands, so I removed this line.

scarlet/source.py
fred3m: It isn't quite that simple, because we also don't create the detection coadd. So I created an `_init_morph` method of `ExtendedSource` that only initializes the morphology, and call that from `PSFDiffKernel._make_initial`.

PR #45, opened 2018-04-25
Multi component

pmelchior:
Add multi-component source models


scarlet/source.py
```diff
@@ -597,3 +707,80 @@ def make_initial(self, img, bg_rms, thresh=1., symmetric=True, monotonic=True, c
             # keep the peak sed
             logger.INFO("Using peak SED for source at {0}/{1}".format(self.center_int[0], self.center_int[1]))
         return sed.reshape((1,B)), morph.reshape((1, morph.shape[0], morph.shape[1]))
+
+
+class MultiComponentSource(ExtendedSource):
```
fred3m: I foresee the need in the future to make spiral galaxies with different stellar populations that *don't* share a common center multi-component sources, this is just a special case where the positions are aligned. We should talk a bit about what's the best way to handle the more general case, when sources aren't aligned, before we put too much work in this class.

scarlet/source.py
```diff
@@ -597,3 +707,80 @@ def make_initial(self, img, bg_rms, thresh=1., symmetric=True, monotonic=True, c
...
+        # create a list of components from morph by layering them on top of each
+        # other and sum up to morph
+        from scipy.ndimage.morphology import binary_erosion
+        K = len(size_percentiles) + 1
+        Ny, Nx = morph.shape[1:]
+        morph_ = np.zeros((K, Ny, Nx))
+        morph_[0,:,:] = morph[0]
+        mask = morph[0] > 0
+        radius = np.sqrt(mask.sum()/np.pi)
+        # make sure they are in decendind order
```
fred3m: descending?

scarlet/source.py
```diff
@@ -15,18 +15,27 @@ class SourceInitError(Exception):
...
 class Source(object):
-
     """A single source in a blend.
 
-    This class is fully functional and acts as base class for specialized
-    initialization, constraints, etc.
+    This class acts as base class for specifying sources and can be
+    extended by specialized initialization, constraints, etc.
     """
-    def __init__(self, sed, morph_image, constraints=None, center=None, psf=None, fix_sed=False,
```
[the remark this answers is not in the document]
fred3m: However, this change broke the docs because some modules specifically pass `morph_image`, so those need to be fixed.

PR #48, opened 2018-06-05
minimal save/load functionality

pmelchior:
dumps all seds/morphologies to numpy `npz` file


fred3m:
Good call, we should have implemented this a long time ago. In the near future we should probably have a way to propagate other features so that the sources and blend can be reconstructed, but this is a good (and useful) start.

pmelchior:
One thing I just realized: this only works if you have the sources still loaded, e.g.
```
blend = scarlet.Blend(sources, ...)
blend.fit()
blend.save("test")
# do something
blend.load("test.npz") # yes, numpy is inconsistent with the filenames...
```
but you can't do `blend = Blend("test.npz")`. What's missing for minimum reload functionality (that is for generating and analyzing models, but not fitting) is the generation of empty sources for each saved component. While we have the sed and morphology, we don't know where the source was in the frame. 

I don't think we need to (or even should) store the PSF, constraints, etc, but we need to be able to generate the idealized scene (pre-convolved in the common sky frame) from the file we save.

That means I will add the center coordinate of the source to the output. This will uniquely define the *resulting* scarlet model.

fred3m:
I think a lot of users will want the ability to recreate the full model, with PSF convolution, and saving `psfs` will allow `Gamma` to be created so that the full model can be realized. This is like adding one more source and shouldn't significantly impact the file size. Saving info about the constraints I see more useful as metadata, so that in the future a user can look back and see what parameters were used to generate the models. This might be easier if we had a simplified interface that could all be stored in the config file to eg. create a constraint from a dictionary of keyword arguments.

pmelchior:
Saving the PSFs is only needed if the user wants to compare the model directly to the original image. I agree that this would be a useful addition, but it means storing an image per band. I'm thinking of saving the content of the sources, with the configuration and observational parameters being left out, for now.

pmelchior:
As a side remark, I think we should create save/load functionality at the level of the source, not the blend. I'll make it such that `Blend.save()` will generate a directory, where the `Source.save()` puts a file. This way the blends are self-contained, we can add blend-level configs and other meta data as needed. But the main purpose is that we can reload a single source if we want to tinker with *another* source in the blend.

pmelchior:
will be dealt with in v0.5 by pickling

PR #52, opened 2018-06-07
Abstract constraints

pmelchior:
This PR modifies `Constraint` to be an abstract implementation, which generates the relevant proximal operators and matrices, which are then subsequently adapted for each source.

This allows the constraints to be shared instead of copied (reducing memory cost and redundant operations) and strongly simplifies the constraint interface.

**Note:** Constraints are now generators of functions, not container of functions! This is necessary to allow the specialization for each source size.

As a bonus: factored out `Cache` to be used consistently in all modules of scarlet.

### DIFF
--- a/scarlet/transformations.py
+++ b/scarlet/transformations.py
@@ -449,7 +449,7 @@ def getSymmetryOp(shape):
     except KeyError:
         idx = np.arange(shape[0]*shape[1])
         sidx = idx[::-1]
-        symmetryOp = getIdentityOp(shape)
+        symmetryOp = getIdentityOp(shape).L
         symmetryOp -= scipy.sparse.coo_matrix((np.ones(size),(idx, sidx)), shape=(size,size))
         symmetryOp = proxmin.utils.MatrixAdapter(symmetryOp, axis=1)
         global cache
--- a/scarlet/transformations.py
+++ b/scarlet/transformations.py
@@ -349,58 +349,13 @@ class LinearOperator(proxmin.utils.MatrixAdapter):
     def __init__(self, L):
         """Initialize the class
         """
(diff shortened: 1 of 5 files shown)


scarlet/transformations.py
```diff
@@ -538,18 +426,18 @@ def getSymmetryOp(shape):
...
         idx = np.arange(shape[0]*shape[1])
         sidx = idx[::-1]
-        symmetryOp = getIdentityOp(shape)
+        symmetryOp = getIdentityOp(shape).L
         symmetryOp -= scipy.sparse.coo_matrix((np.ones(size),(idx, sidx)), shape=(size,size))
-        symmetryOp = LinearOperator(symmetryOp)
-        global cache
-        cache[name][key] = symmetryOp
+        symmetryOp = proxmin.utils.MatrixAdapter(symmetryOp, axis=1)
+        _ = symmetryOp.spectral_norm
```
fred3m: I think you don't need the `_ = ` here, and this can just be
```
symmetryOp.spectral_norm
```
to calculate the spectral norm.

scarlet/transformations.py
```diff
@@ -746,10 +633,9 @@ def getRadialMonotonicOp(shape, useNearest=True, minGradient=1, subtract=True):
             monotonic = cosArr-scipy.sparse.diags(diagonal, offsets=0)
         else:
             monotonic = cosArr
-        monotonic = LinearOperator(monotonic.tocoo())
-
-        global cache
-        cache[name][key] = monotonic
+        monotonic = proxmin.utils.MatrixAdapter(monotonic.tocoo(), axis=1)
+        _ = monotonic.spectral_norm
```
fred3m: same here

PR #57, opened 2018-07-17
acceleration as config option

pmelchior:
also: algorithm calls are now explicit

does not require proxmin update since the algorithm decision is made in scarlet.


pmelchior:
Yes, I thought about it, but it's been improper before: you ask for one algorithm and then internally your call gets rerouted to another one. Sure, it's faster, but it's also not what you requested.

PR #58, opened 2018-07-27
Bug fixes

fred3m:


fred3m:
@pmelchior would you mind reviewing this so I can merge it? I just fixed a couple of bugs:

1. Somewhere along the way the wrong parameter was used for `max_iter` in `Blend.fit`

1. I added a check so that if `bg_rms=0` the init method of `ExtendedSource` does not crash.

PR #61, opened 2018-08-06
convergence flag; closes #60

### LINKED ISSUE #60: Set flag when fit hasn't converged
Currently there is only a warning printed by proxmin when the fit hasn't converged. It would be useful to set an internal flag in such cases.

pmelchior:
Sets a flag if the `fit` method has converged within the specified `e_rel`. Requires a proxmin update.

### DIFF
--- a/scarlet/blend.py
+++ b/scarlet/blend.py
@@ -116,6 +116,7 @@ class Blend(ComponentTree):
         except AttributeError:
             self.it = 0
             self._model_it = -1
+            self.converged = False
             # Caches for 1/Lipschitz for A and S
             self._cbAS = [proxmin.utils.ApproximateCache(self._one_over_lipschitz, slack=self.config.slack),
                           proxmin.utils.ApproximateCache(self._one_over_lipschitz, slack=self.config.slack)]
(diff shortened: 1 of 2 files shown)


scarlet/blend.py
```diff
@@ -172,6 +173,9 @@ def fit(self, steps=200, e_rel=1e-2):
                     Ls=self._Ls, update=update, update_order=update_order, steps_g_update=steps_g_update, max_iter=steps,
                     e_rel=self._e_rel, e_abs=self._e_abs)
 
+            X, converged, errors = res
+            self.converged = all(converged)
```
fred3m: It would probably be useful to also store the convergence of each component, so `self.convergence=converged`. This might make it easier to troubleshoot in the future.

pmelchior: In principle yes, but the ordering of that structure is not obvious outside of `Blend.fit`: it depends on the `update_order` keyword if its all components of `A` and then of `S` or the other way round.

I expressly don't want to make it a property of `Component` even though what you suggest is a per-component flag.

pmelchior: that'll work. I'll unpack the convergence flags in `fit` and add a `has_converged(k)` method which spits out the a tuple for `(A,S)`.

pmelchior: even better: I made `self.converged` a list of those tuples, ordered by `k`. It's very close to your original request and avoid the new method

PR #64, opened 2018-09-26
Crowded fields

fred3m:
This pull request implements features that will be useful in crowded fields, including using point sources with fixed (single pixel) morphologies and normalization of the morphology (as opposed to the SED).

### DIFF
--- a/scarlet/config.py
+++ b/scarlet/config.py
@@ -1,4 +1,13 @@
+from enum import Enum
+
 import numpy as np
+
+
+class Normalization(Enum):
+    A = 1
+    S = 2
+    Smax = 3
+
 
 class Config(object):
     """Blend Configuration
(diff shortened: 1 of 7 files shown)


scarlet/config.py
```diff
@@ -1,5 +1,14 @@
+from enum import Enum
+
 import numpy as np
 
+
+class Normalization(Enum):
```
pmelchior: Could you add docstrings, please

scarlet/config.py
```diff
@@ -31,7 +40,8 @@ class Config(object):
         (`exact_lipschitz` is `False`)
     """
     def __init__(self, accelerated=True, update_order=None, slack=0.2, refine_skip=10, source_sizes=None,
-                 center_min_dist=1e-3, edge_flux_thresh=1., exact_lipschitz=False):
+                 center_min_dist=1e-3, edge_flux_thresh=1., exact_lipschitz=False,
+                 normalization=Normalization.A, update_model=False):
```
pmelchior: same here: docstrings

pmelchior:
I'm primarily worried about the conceptual structure. The normalization is a constraint. We treat the constraints on a per-component basis. Why should the normalization be specified differently?

I take your point that for bulk LSST processing, we wouldn't mix A and S normalizations. But you might want to do so for special (e.g. science platform) queries. I don't won't to preclude those.

pmelchior:
I don't think that's necessary if we provide a clear default. People would have to intentionally overwrite it

pmelchior:
Not a part of this PR, but we should clarify when the different normalization options are better. Our initial notion was the S normalization is normally faster to converge and yields a A matrix that is simpler to interpret.

Issue #71, opened 2019-02-07
Normalization is a Constraint


pmelchior:
PR #64  introduced normalization options. Technically these are just another type of constraint, but they are currently handled differently. This issue is meant to correct that in the minimally invasive way.

The normalization is currently only used in the Source initialization, and it only changes the behavior of `SimpleConstraint`, the only thing that needs to be done is to remove the normalization class member of `Source`.

Issue #70, opened 2019-02-07
Flux-dependent operations in sources don't know about normalization


pmelchior:
So far, we've always assumed the A is normalized and the we get the flux of a component by summing of S. That's not necessarily true anymore. Problems arise e.g. here:
https://github.com/fred3m/scarlet/blob/master/scarlet/source.py#L62

We need to provide a per-component method, e.g. `get_flux()` that is aware of the normalization. This is a problem with the current scheme because the constraints are held by the components, but the normalization is held by the sources. Operationally both are constraints and should be treated equally, in which case both are held by the components. It's true that it's unlikely that someone would ever mix two components with different normalizations, but we've decided to retain that option.

So, this issue is for
* creating `Component.get_flux()`
* checking that we consistently call this method whenever the component flux is required.

This leaves one problem: what do we mean by flux? I argue it needs to be bolometric, i.e. `flux = morph.sum() * sed.sum()`. This form is -- yeah -- independent of normalization, which simplifies the problem considerably: components don't need to carry the normalization around at all.

Another issue will address the "normalization is a constraint problem".

fred3m:
You make a good point. My only comment is that `get_flux` should not return a bolometric flux, but an array with the flux in each band, ie `flux=morph.sum() * sed`. The user can make it bolometric if they want, but this allows for more flexibility in cases (like the one you link to in your comment) where we want to weight sources by the flux in each band. We should also add `get_flux` to `Source`, where calling it returns the sum of `get_flux` for each component.

PR #72, opened 2019-02-07
Normalization

pmelchior:
This PR provides a consistent treatment of normalization at the component level, following the proposals of #70 and #71

### DIFF
--- a/scarlet/source.py
+++ b/scarlet/source.py
@@ -165,7 +165,6 @@ class PointSource(Source):
             config = Config()
         if shape is None:
             shape = (config.source_sizes[0],) * 2
-        self.normalization = normalization
         sed, morph = self._make_initial(center, img, shape, psf, config, tiny)
 
         if constraints is None:
@@ -175,6 +174,7 @@ class PointSource(Source):
 
         component = Component(sed, morph, center=center, constraints=constraints, psf=psf, fix_sed=fix_sed,
                               fix_morph=fix_morph, fix_frame=fix_frame, shift_center=shift_center)
+        component.normalize(normalization)
         super(PointSource, self).__init__(component)
 
     def _make_initial(self, center, img, shape, psf, config, tiny=1e-10):
@@ -191,32 +191,16 @@ class PointSource(Source):
         """
         B, Ny, Nx = img.shape
         _y, _x = center_int = np.round(center).astype('int')
-        if self.normalization != sc.Normalization.A:
-            sed = img[:, _y, _x]
-            if psf is not None:
-                # Increase the magnitude of the SED's to account
-                # for the normalized PSF's
-                py, px = (psf.shape[0] // 2, psf.shape[1] //2)
-                centers = psf[:, py, px]
-                centers = np.max(psf, axis=(1,2))
-                sed = sed/centers
-            sed = prox_sed_on(sed, 0)
-            morph = np.zeros(shape, dtype=img.dtype)
-            # Turn on a single pixel at the peak
-            cy, cx = (shape[0] // 2, shape[1] //2)
-            morph[cy, cx] = 1
-            return sed, morph
-        else:
-            # determine initial SED from peak position
-            try:
-                sed = get_pixel_sed(img, center_int)
-            except SourceInitError:
-                # flat weights as fall-back
-                sed = np.ones(B) / B
-            morph = np.zeros(shape)
-            # Turn on a single pixel at the peak
-            cy, cx = (shape[0] // 2, shape[1] //2)
-            morph[cy, cx] = max(img[:,_y,_x].sum(axis=0), tiny)
+        # determine initial SED from peak position
+        try:
+            sed = get_pixel_sed(img, center_int)
+        except SourceInitError:
+            # flat weights as fall-back
+            sed = np.ones(B) / B
+        morph = np.zeros(shape)
+        # Turn on a single pixel at the peak
+        cy, cx = (shape[0] // 2, shape[1] //2)
+        morph[cy, cx] = max(img[:,_y,_x].sum(axis=0), tiny)
         return sed, morph
 
 class ExtendedSource(Source):
@@ -260,7 +244,6 @@ class ExtendedSource(Source):
         if config is None:
             config = Config()
 
-        self.normalization = normalization
         sed, morph = self._make_initial(center, img, bg_rms, thresh=thresh, symmetric=symmetric,
                                         monotonic=monotonic, config=config)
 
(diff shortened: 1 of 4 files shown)


scarlet/source.py
```diff
@@ -53,13 +53,23 @@ def get_model(self):
...
         NOTE: Only components with `shift_center > 0` will be moved.
         """
         if len(self.components) > 1:
-            _flux = np.array([c.morph.sum() for c in self.components])
+            # TODO: find optimal centering Weights
+            # With flux weighting, larger components often have large weights
+            # even if they are shallow and thus hard to center ...
+            # _flux = np.array([c.get_flux().sum() for c in self.components])
+            #  ... use flat weights instead
+            _flux = np.array([1 for c in self.components])
```
fred3m: I agree that the older method did not necessarily have the desired affect. Perhaps if we just use a small window (3px x 3 px) we can weight based on the _central_ flux for each component. This should give more weight to components with more flux concentrated at the center, which should be easier to fit. Including the new parameter `center_window` that can either be an array or window size initialized as follows:
```
if np.isscalar(center_window):
    center_window = np.ones((center_window, center_window), dtype=float)
```

This method then becomes:
```
_flux = np.array([np.sum(c.morph[cy-ry:cy+ry+1, cx-rx:cx+rx+1] * self.center_window) for c in self.components])
```
where cy,cx is the location of the central pixel in the morphology and ry, rx is the size of the window.

This allows the user to pass an array like `center_window=np.array([[0,1,0],[1,1,1],[0,1,0]])` to choose the exact window used.

If you agree I can open an issue to make this change.

scarlet/source.py
```diff
@@ -191,32 +201,16 @@ def _make_initial(self, center, img, shape, psf, config, tiny=1e-10):
...
+        # determine initial SED from peak position
+        try:
+            sed = get_pixel_sed(img, center_int)
+        except SourceInitError:
+            # flat weights as fall-back
+            sed = np.ones(B) / B
+        morph = np.zeros(shape)
+        # Turn on a single pixel at the peak
+        cy, cx = (shape[0] // 2, shape[1] //2)
+        morph[cy, cx] = max(img[:,_y,_x].sum(axis=0), tiny)
```
fred3m: I don't like this because it won't initialize a source with S normalization properly.

scarlet/source.py
```diff
@@ -309,18 +303,6 @@ def _make_initial(self, center, img, bg_rms, thresh=1., symmetric=True, monotoni
...
-        # For the different S matrix normalizations,
-        # we make sure that our inital values are in agreement
-        # with the correct normalization
-        if self.normalization != sc.Normalization.A:
-            if self.normalization == sc.Normalization.S:
-                norm = np.sum(morph)
-            elif self.normalization == sc.Normalization.Smax:
-                norm = np.max(morph)
-            sed *= norm
-            morph /= norm
```
fred3m: Again, this will not initialize a source using S normalization properly

scarlet/source.py
pmelchior: My reply refers to the normalization issue. It's actually correct. There's a `component.normalize` call in *every* source type, which takes care of the different normalizations. The problem is in the `PointSource` initialization only.

scarlet/source.py
pmelchior: yes, it will. Check `ExtendedSource`

pmelchior: OK, I see your point now. Let me think...

pmelchior:
I've reworked the initialization so that it makes more sense. The default setup is Smax because then you can simple start with the SED of the brightest pixel, as we always did. As a consequence, this SED is not normalized, it's simply the values of that pixel. From there on the desired normalization is computed for each component by `Component._normalize()`.

I've made sure that point, extended and multi-component sources are now initialized correctly in every desired normalization option.

scarlet/source.py
```diff
@@ -99,24 +109,21 @@ def get_pixel_sed(img, position):
         # which will cause the code to crash later
         msg = "Zero or negative flux at y={0}, x={1}"
         raise SourceInitError(msg.format(_y, _x))
+    return sed
 
-    # ensure proper normalization
-    return proxmin.operators.prox_unity_plus(sed, 0)
-
-def get_integrated_sed(img, weight):
+def get_integrated_sed(img, weight, p=1):
```
fred3m: please update the docstring and use a more descriptive variable name than `p`.

scarlet/source.py
```diff
@@ -175,6 +181,7 @@ def __init__(self, center, img, shape=None, constraints=None, psf=None, config=N
 
         component = Component(sed, morph, center=center, constraints=constraints, psf=psf, fix_sed=fix_sed,
                               fix_morph=fix_morph, fix_frame=fix_frame, shift_center=shift_center)
+        component._normalize(normalization)
```
fred3m: I like this method. It seems like a good way to separate the constraints.

scarlet/source.py
```diff
@@ -191,32 +198,12 @@ def _make_initial(self, center, img, shape, psf, config, tiny=1e-10):
...
-            except SourceInitError:
-                # flat weights as fall-back
-                sed = np.ones(B) / B
-            morph = np.zeros(shape)
-            # Turn on a single pixel at the peak
-            cy, cx = (shape[0] // 2, shape[1] //2)
-            morph[cy, cx] = max(img[:,_y,_x].sum(axis=0), tiny)
+        # determine initial SED from peak position: amplitude is in sed
+        sed = get_pixel_sed(img, center_int)
+        morph = np.zeros(shape)
```
fred3m: This should be `morph = np.zeros(shape, dtype=img.dtype)`. This is one part of the code that we need to clean up because the stack often uses 32 bit floats instead of doubles and at some point we'll need to go back through the code and find everywhere that we don't set the type.

pmelchior:
@fred3m merge if you have no objections

fred3m:
It still fails to build the docs for `user_docs.ipynb`. Can you test that out and see if it runs properly on your machine? It fails on both RTD and when I run just that notebook locally. I think that the `BuldgeDisk` example of a custom source needs to be updated with the new normalization scheme.

pmelchior:
Hmm, got the sucker. It's not a bug, it's yet another acceleration problem. `accelerated=False` does the trick. We *really* need to switch it off.

pmelchior:
Btw, `Normalization.S` is unstable for the bulge-disk case, `Smax` works like charm.

fred3m:
Looks good, feel free to merge. I have a ticket out that I'll finish this weekend to max Smax and no acceleration the default, and update the docs accordingly.

PR #74, opened 2019-02-15
tickets/DM-17847: Update defaults in scarlet and create changelog

fred3m:
I also switched from 0.3 to 0.4. In the future we should probably switch subversions more often, as this one hadn't been changed in almost 9 months. I did my best making the change-log since that version, with more focus on the changes made recently, since most people seem to be using the latest master anyway. In the future we should include a change log entry with each PR.

### DIFF
--- a/changes.rst
+++ b/changes.rst
@@ -1,0 +1,53 @@
+0.4 (unreleased)
+----------------
+
+General
+^^^^^^^
+ - Dropped python 2 support
+
+ New Features
+ ^^^^^^^^^^^^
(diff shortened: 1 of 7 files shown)


pmelchior:
I see no problems with it, but can you move the change log to the docs and include it at the top level so that it appears on the left-hand side navigation. I think it's better placed there than in the main code directory

fred3m:
Ok, I moved the change log to the docs. Is it ok to merge now?

PR #75, opened 2019-02-20
tickets/DM-17960

fred3m:
Add code to do PSF convolutions and translations using FFT's. This does not include the ability to use these during optimization, just the functionality to perform convolutions using FFT's.


pmelchior:
I have no problems with the fft and resampling code, but the testing in the docs isn't quite right:
`test_resampling_algorithm` uses a sampled Gaussian as truth, which ignores the pixelation. To do this correctly, you have to integrate the Gaussian within the pixels (either analytically or numerically by oversampling). Otherwise the effects of pixelation aren't correctly treated, which will makes it hard to interpret the interpolation errors.

fred3m:
That's a very good point. I'll revisit this today.

pmelchior:
This PR has become obsolete by #77

PR #77, opened 2019-03-05
Create tests to run in Travis CI

fred3m:


pmelchior:
This is a combined PR  for new implementation of resampling kernels and convolutions with FFTs *plus* the new testing framework Travis.

This will provide the methods for the next major revision that will increase performance and make the code more modular, which we'll need for multi-instrument support.

PR #83, opened 2019-05-08
Switch from pytorch to autograd

fred3m:
In order to reduce the dependencies of scarlet and make it easier to use the code has been modified to switch from using `pytorch` to the `autograd` package to calculate gradients.


scarlet/component.py
```diff
@@ -123,31 +123,37 @@ def coord(self):
...
 
     @property
     def morph(self):
         """Numpy view of the component morphology
         """
-        return self._morph.data.detach().numpy()
+        return self._morph
 
-    def get_model(self, numpy=True):
+    def get_model(self, sed=None, morph=None):
```
pmelchior: When do you need the ability to call this with some other sed and morphology?

scarlet/component.py
```diff
@@ -286,25 +292,32 @@ def coord(self):
             else:
                 return (self._index,)
 
-    def get_model(self, numpy=True):
+    def get_model(self, seds=None, morphs=None):
```
pmelchior: same question as above

scarlet/component.py
fred3m: This is what I was talking about today with the way autograd calculates gradients. You basically need to setup `nabla = grad(loss_function, (a long tuple here))` to calculate the partial derivatives, and call with with `nabla(sed1, sed2, ..., morph1, morph2, ...)`. So inside the loss function you have to build the model with the seds's and morphologies passed to the loss function. See https://github.com/fred3m/scarlet/blob/auto/scarlet/blend.py#L111-L146. 

This is different than pytorch, since it knew which parameters to use for the partial derivatives.

pmelchior: I understand that `nabla` requires all parameters, but `blend.get_model` should work identically without arguments. *I think* you only need to declare the parameters for autograd, but you don't need to strictly pass them.

pmelchior: In essence, replace https://github.com/fred3m/scarlet/blob/auto/scarlet/blend.py#L136-L140 with
```
model = self.get_model()
```

fred3m: Unfortunately no. Inside the loss function I inserted
```
print(type(seds[0]))
print(type(self.sources[0].sed))
```
at https://github.com/fred3m/scarlet/blob/auto/scarlet/blend.py#L139 and the output was

```
<class 'autograd.numpy.numpy_boxes.ArrayBox'>
<class 'numpy.ndarray'>
```

What happens is the parameters passed to `nabla` are wrapped in the `ArrayBox` class, which is what autograd uses to propagate the gradients. Architecturally this is very different than the `Jax` and `pytorch` implementations.

pmelchior: Understood, thanks for checking

pmelchior:
@herjy will not be happy, since his pytorch implementation was essentially ready. But I'm happy that we could cut down on dependencies

fred3m:
Sorry @herjy . 🙁 Fortunately it shouldn't require significant changes, since to update the code on my end I basically just swapped out `torch.rfft...` with `np.fft.fft2...` in observation and it pretty much just worked (once I made the underlying changes everywhere else).

herjy:
I'll survive, plus I already moved my part of the code to another class, so not much to change.

PR #84, opened 2019-05-13
Make models more general

fred3m:
Components can now use any number of parameters and are no longer restricted to just `sed` and `morph`. This requires some structural changes to allow them to update their gradients, moving the majority of `Blend._backward` into `Component`, including the actual gradient update.

A bug fix was also implemented (`interpolation.fft_convolve` used the `numpy.prod` function that was not wrapped in `autograd`) and dead code (`the symmetric center algorithm`) was stripped out.


scarlet/component.py
```diff
@@ -292,31 +317,39 @@ def coord(self):
...
         """
         model = np.zeros((self.B, self.Ny, self.Nx))
-        for k in range(self.K):
-            if seds is not None and morphs is not None:
-                _model = self.components[k].get_model(seds[k], morphs[k])
+        pidx = 0
+        if len(parameters) > 0:
+            for k in range(self.K):
+                p = len(self.components[k].parameters)
+                _model = self.components[k].get_model(*parameters[pidx:pidx+p])
```
pmelchior: I suggest `model += self.components[k].get_model(*parameters[pidx:pidx+p])` here

scarlet/component.py
```diff
@@ -292,31 +317,39 @@ def coord(self):
...
+                pidx += p
                 model = model + _model
-            else:
+        else:
+            for k in range(self.K):
                 model = model + self.components[k].get_model()
```
pmelchior: same here

scarlet/component.py
```diff
@@ -87,8 +87,8 @@ def __init__(self, sed, morph, prior=None, fix_sed=False, fix_morph=False):
         # set sed and morph
         self._sed = np.array(sed)
```
pmelchior: I think we can expose sed and morph again since they're are ordinary numpy arrays now

scarlet/component.py
```diff
@@ -155,6 +167,10 @@ def get_model(self, sed=None, morph=None):
             return self._sed[:, None, None] * self._morph[None, :, :]
         raise ValueError("You need to supply `sed` and `morph` or neither")
 
+    def update_gradients(self, grad_sed, grad_morph):
```
pmelchior: I don't get this function: it has a fixed number of specified parameters. isn't the idea of this PR to allow for an arbitrary arrangement of parameters.

scarlet/component.py
fred3m: You can't use an inline update with autograd, because it doesn't know that you made a change. See https://github.com/HIPS/autograd/blob/master/docs/tutorial.md#dont-use

scarlet/component.py
fred3m: I like hiding them so that users don't try to modify them in place and get unexpected results.

scarlet/component.py
fred3m: This is because most components will just have `sed` and `morph`. For sources with more complicated models `get_model` is actually rewritten by an inherited class (as it must be, because it will contain a different model). I'll post an example on slack.

pmelchior:
Can we close this PR? It's not an active development, and I can delete the `fullscene` branch without closing this PR. Just to be clear, I don't want to delete the `parameters` branch, which has independent code development, just `fullscene`, which is fully merged

fred3m:
Yes, that's fine. I'll close the PR but keep the branch.

PR #86, opened 2019-06-03
Post sprint

fred3m:


scarlet/bbox.py
```diff
@@ -0,0 +1,171 @@
+import autograd.numpy as np
+
+
+class BoundingBox(object):
```
pmelchior: Is this class actually used anywhere?

scarlet/observation.py
```diff
@@ -0,0 +1,233 @@
...
+        Returns
+        -------
+        model: array
+            The convolved and resampled `model` in the observation frame.
+        """
+        if self.wcs is not None:
+            msg = "get_model is currently only supported when the observation frame matches the scene"
+            raise NotImplementedError(msg)
+
+        def _convolve_band(model, psf_fft):
```
pmelchior: Is there a reason why this is not an ordinary class method?

scarlet/observation.py
```diff
@@ -0,0 +1,233 @@
...
+
+        Returns
+        -------
+        result: array
+            Scalar tensor with the likelihood of the model
+            given the image data.
+        """
+        if self._psfs is not None:
+            model = self.get_model(model)
+        return np.sum(0.5 * (self._weights * (model - self._images))**2)
```
pmelchior: better `0.5 * np.sum(...)`: one op instead of B*Ny*Nx

scarlet/observation.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def get_scene(self, scene):
```
pmelchior: The utility of this method is unclear to me. I know that we talked about its potential uses e.g. to initialize sources, but our conclusion was that it cannot serve its purpose as one would have to do deconvolutions etc

I suggest to remove

scarlet/bbox.py
fred3m: Yes, monotonicity uses it to save CPU time and only apply monotonicity to the bounding box. Symmetry uses it as well for the same reason.

scarlet/source.py
```diff
@@ -1,396 +1,338 @@
...
-            detect = np.einsum('i,i...', weights, img) / jacobian
-
-            # thresh is multiple above the rms of detect (weighted variance across bands)
-            bg_cutoff = thresh * np.sqrt((weights**2 * bg_rms**2).sum()) / jacobian
-        else:
-            detect = np.sum(img, axis=0)
-            bg_cutoff = 0
-        morph = self._init_morph(detect, center, bg_cutoff, symmetric, monotonic, config)
+        sed, morph = init_extended_source(sky_coord, scene, observations, bg_rms, obs_idx,
+                                          thresh, True, monotonic)
```
pmelchior: It's a bit weird that the initialization is always symmetric even if the source itself may not have to. I think this isn't transparent to the user.

scarlet/update.py
```diff
@@ -0,0 +1,221 @@
...
+        norm = component.morph.sum()
+        component.sed[:] = component.sed * norm
+        component.morph[:] = component.morph / norm
+    elif t == 'morph_max':
+        norm = component.morph.max()
+        component.sed[:] = component.sed * norm
+        component.morph[:] = component.morph / norm
+    else:
+        raise ValueError("Unrecognized normalization '{0}'".format(type))
+
```
pmelchior: The method needs to `return component`

scarlet/source.py
fred3m: I get that, but this really helps initialize compact sources and makes a big difference in the convergence time. Since we expect users looking at more complicated objects to override the initialization anyway, I'm fine with hiding this under the hood.

scarlet/observation.py
fred3m: I suppose that it could be. It's never used outside of `get_model` so I didn't see the need to make it available outside the method, but I can change that.

pmelchior: it's inside of a method that gets called multiple times, so the interpreter has to run every time. Plus, it's easier to inherit from it when the method is available at the class level

fred3m: good point

scarlet/observation.py
fred3m: removed

PR #87, opened 2019-06-05
g Reconvampling

herjy:
Resampling and convolution accounted for for datasets at different resolutions. 
This includes a set of functions in 'interpolation.py' and 'resampling.py' along with a new class 'Combination' that inherits from 'Scene' and new initialisation functions.


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,14 @@
+<component name="InspectionProjectProfileManager">
```
pmelchior: I'm pretty sure that this file should not be part of the code base. Please remove

docs/component.ipynb
```diff
@@ -22,11 +22,7 @@
   },
   {
    "cell_type": "raw",
-   "metadata": {
-    "deletable": true,
-    "editable": true,
-    "raw_mimetype": "text/restructuredtext"
-   },
+   "metadata": {},
```
pmelchior: It looks like you've committed all doc notebooks after you executed them locally. Did you actually made edits to the docs?

scarlet/blend.py
```diff
@@ -8,7 +8,7 @@
 logger = logging.getLogger("scarlet.blend")
 
 
-class Blend(ComponentTree, Scene):
+class Blend(ComponentTree):
```
pmelchior: This is now somewhat inconsistent:
* `Blend` is now not derived from `Scene` anymore, but line 32 below will attempt to initialize `self` as `Scene`. I'm surprised that this even works.
* `Observation` is still derived from `Scene`, so now `Blend` and `Observation` behave differently in this regard: the former *has* a scene, the latter *is* a scene.

We have agreed that we would alter this consistently, i.e. both should *have* `MetaData` because that clarifies what we mean by `Scene`. For now, please derive `Blend` from `ComponentTree, Scene` as before.

scarlet/component.py
```diff
@@ -263,7 +263,16 @@ def B(self):
         """Number of bands
         """
         return self.components[0].B
-
+    @property
```
pmelchior: Why are those needed? You should not need to talk to the components directly as the resampling only operates on the full scene

scarlet/interpolation.py
```diff
@@ -295,6 +293,26 @@ def get_separable_kernel(dy, dx, kernel=lanczos, **kwargs):
     kyx = np.outer(ky, kx)
     return kyx, y_window, x_window
 
+def sinc_interp(coord_hr, coord_lr, Fm):
+    '''
+    INPUTS:
```
pmelchior: doc string convention: `INPUTS:` -> `Parameters` (no colon)

scarlet/interpolation.py
```diff
@@ -295,6 +293,26 @@ def get_separable_kernel(dy, dx, kernel=lanczos, **kwargs):
...
+    ------
+        coord_hr: Coordinates of the high resolution grid
+        coord_lr: Coordinates of the low resolution grid
+        Fm: Sample at positions coord_lr
+    OUTPUTS:
```
pmelchior: doc string convention: `OUTPUTS:` -> `Returns`

scarlet/interpolation.py
```diff
@@ -295,6 +293,26 @@ def get_separable_kernel(dy, dx, kernel=lanczos, **kwargs):
...
+        result:  interpolated  samples at positions coord_hr
+    '''
+    a, b = coord_hr
+    A, B = coord_lr
+    hx = np.abs(A[1]-A[0])
+    hy = np.abs(B[np.int(np.sqrt(np.size(B)))+1] - B[0])
+
+    assert hx != 0
+
+    return np.array([Fm[k] * sinc2D((a-A[k])/(hx),(b-B[k])/(hy)) for k in range(len(A))]).sum(axis=0)
```
pmelchior: Can you avoid the for loop by writing `sinc2D` to operate on arrays?

scarlet/observation.py
```diff
@@ -146,11 +152,32 @@ def match(self, scene):
                 new_kernel_fft.append(np.fft.fft2(np.fft.ifftshift(kernel)))
             self.psfs_fft = np.array(new_kernel_fft)
 
-        # 4) divide obs.psf from scene.psf in Fourier space
+        return self
 
-        # [ 5) compute sparse representation of interpolation * convolution ]
+    @property
```
pmelchior: Again, why are these needed? `Observation` *is* a `Scene`, so these methods are fully redundant. Even if we change it so that `Observation` has a `Scene`, you can always call these methods of the class member `scene`. Please remove

scarlet/observation.py
```diff
@@ -146,11 +152,32 @@ def match(self, scene):
...
+        """Number of pixels in the x-direction
+        """
+        return self.shape[2]
 
     @property
+    def shape(self):
+        """Shape of the model.
+        """
+        return self._shape
+    @property
```
pmelchior: I remain skeptical that we need to hide `images` at all. This is a left-over from pytorch, where they had different type, but I think we can remove that. @fred3m ?

.idea/inspectionProfiles/Project_Default.xml
herjy: Yup, Done!

docs/component.ipynb
herjy: It looks like it although I don't remember executing this particular notebook. I sure did not change anything in there. 
Is there anything I should do about that?

scarlet/blend.py
herjy: Well, I am happy to revert this back, since apparently I missed the concept, but I actually thought blend as an object wrapped around the models while Scene and Observations are representations of the space in which the data live. That helped me disambiguating between get_model in observations and in Blend.

herjy: Having just said that, I realise that Scene is the abstract space between the data and the model and as such, it makes sense that Blend and Observations both inherit from Scene. Fine I'll do the change.

scarlet/component.py
herjy: Old stuff that I thought I neede. 
Deleted

scarlet/interpolation.py
herjy: The problem is not the sinc. a is an array. It's the dimensions of a and A that do not match and made me use the loop.

herjy: Once again, as I say this, I realise there is a smart solution that uses newaxis to shunt the for loop. It works, but I want to confirm that tomorrow on my new machine where I am perfectly sure of the setup.

scarlet/observation.py
herjy: same as before

scarlet/observation.py
herjy: Same, it's an old thing I used for testing or possibly when I very briefly pytorched my code.

fred3m: This is a style issue. Because we initialize the observation with `images` and it is used to set the shape `(B, Ny, Nx)`, which also sets the image and PSF padding, having a user change `images` can have unexpected consequences that are difficult to diagnose. So I like to use a more C++ style approach and hide any variables that we really don't want the user to set, just to be safe.

Your argument is usually that the user should know better and they get what they deserve. I try to be a bit more forgiving *and* try to save us from emails from confused users with hard to diagnose issues. In my mind you only get what you deserve when you go around and monkey with hidden variables...

scarlet/blend.py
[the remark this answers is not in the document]
fred3m: I see you already touched on this above. Resolving this conversation

docs/component.ipynb
herjy: No edits, no

pmelchior: Ok, then reset the docs files with `git checkout -- filename`

scarlet/observation.py
```diff
@@ -231,3 +253,146 @@ def get_scene(self, scene):
             msg = "get_scene is currently only supported when the observation frame matches the scene"
             raise NotImplementedError(msg)
         return self.images
+
+
+
+class Combination(Scene):
```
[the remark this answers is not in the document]
pmelchior: The opposite. I suggest `LowResObservation`. We may have to think about the nomenclature again when we also have high spectral resolution, but for the time being that makes it clear that this observation has lower resolution than `Blend`

scarlet/observation.py
```diff
@@ -231,3 +253,146 @@ def get_scene(self, scene):
...
+                _target = self.target_psf
+            # 3) compute obs.psf in the frame of scene
+            resconv_op = []
+
+            for _psf in self.psfs:
+
+
+                #Computes spatially matching observation and target psfs. The observation psf is also resampled to the scene's resolution
+
+                new_target, observed_psf = resampling.match_psfs(_target, _psf, whr, wlr)
```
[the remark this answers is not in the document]
pmelchior: I'm also not following what's going on here

scarlet/observation.py
```diff
@@ -231,3 +253,146 @@ def get_scene(self, scene):
...
+        img[:, self._over_lr[0], self._over_lr[1]] = self.get_model(model)
+        return img
+
+    def get_loss(self, model):
+        if self._psfs is not None:
+            model = self.get_model(model)
+
+        return np.sum(0.5 * (self._weights * (model - self.images[:,self._over_lr[0].astype(int),self._over_lr[1].astype(int)]))**2)
+
+    def get_scene(self, image):
```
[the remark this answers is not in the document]
pmelchior: Sure, but this method doesn't fully work as advertised as there's no PSF deconvolution. Both of you have to tell me what this method is supposed to do and when that is needed

scarlet/resampling.py
```diff
@@ -0,0 +1,239 @@
...
+    for m in range(np.size(B)):
+            mat[:, m] = conv2D_fft(shape, A[m], B[m], p, h, scp).flatten()*A.size/a.size
+            #mat[:, m] /= np.sum(mat[:,m])
+
+    return mat
+
+
+
+
+def linorm2D(S, nit):
```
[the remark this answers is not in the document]
pmelchior: That's the [power iteration](https://en.wikipedia.org/wiki/Power_iteration)

scarlet/resampling.py
```diff
@@ -0,0 +1,239 @@
...
+    # Mask of low resolution pixels in the overlap at low resolution:
+    over_lr = ((Y_hr>0) * (Y_hr<n2) * (X_hr>0) * (X_hr<n1))
+    # Mask of low resolution pixels in the overlap at high resolution:
+    over_hr = ((y_lr>0) * (y_lr<N2) * (x_lr>0) * (x_lr<N1))
+
+
+    mask = over_hr.reshape(n1,n2)
+
+    if np.sum(mask) == 0:
+        raise ValueError('No overlap found between datasets. Check your coordinates and/or WCSs.')
```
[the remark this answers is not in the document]
pmelchior: Simply define your own exception by deriving from Exception, e.g.
```
class SourceInitError(Exception):
    """Error during source initialization
    """
    pass
```

scarlet/source.py
```diff
@@ -343,6 +472,50 @@ def __init__(self, sky_coord, scene, observations, bg_rms, obs_idx=0, thresh=1,
         Component.__init__(self, sed, morph, **component_kwargs)
 
 
+class Extended_CombinedSource(PointSource):
```
pmelchior: Let's call this `CombinedExtendedSource`

scarlet/blend.py
```diff
@@ -36,8 +36,12 @@ def __init__(self, scene, sources, observations):
         except TypeError:
             observations = (observations,)
         self.observations = observations
+        self.scene = scene
+
+
```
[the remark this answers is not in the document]
herjy: For all of these, I used an automatic formating this time. It should be fine, but let's make sure of it

docs/component.ipynb
herjy: I came back 8 steps on those

scarlet/observation.py
```diff
@@ -67,19 +67,22 @@ def psfs(self):
...
         """
         if self.wcs is not None:
-            return self.wcs.radec2pix(sky_coord).int()
-        return (int(sky_coord[0]), int(sky_coord[1]))
+            if np.size(self.wcs.array_shape) == 3:
+                coord = self.wcs.wcs_world2pix(sky_coord[0], sky_coord[1], 0, 0)
+            elif np.size(self.wcs.array_shape) == 2:
+                coord = self.wcs.wcs_world2pix(sky_coord[0], sky_coord[1], 0)
+
+        return (coord[0].item(), coord[1].item())
```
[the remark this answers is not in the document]
herjy: astropy is doing (y,x)

scarlet/observation.py
```diff
@@ -231,3 +253,146 @@ def get_scene(self, scene):
...
+            whr = scene.wcs
+            wlr = self.wcs
+
+            #Reference PSF
+            if self.target_psf is None:
+
+                if np.size(np.shape(scene.psfs)) ==2:
+                    _target = scene.psfs
+                    _shape = scene.shape
+                elif np.size(np.shape(scene.psfs)) ==3:
```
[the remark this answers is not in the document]
herjy: Always 3-D is the best option I think

scarlet/resampling.py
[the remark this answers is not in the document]
herjy: Yup, sorry, lazy writing

scarlet/observation.py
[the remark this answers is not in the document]
herjy: I called it ObservationToResample for now. Pretty explicit.
The problem with LowResObservation is that we might want to match observations at the same grid resolution from different telescopes, in which case this hierarchical nomenclature can be confusing. 
Saying this, I realise that many variable are named based on this hierarchical conception that low resolution data are matched to higher resolution (x_lr, x_hr).

scarlet/observation.py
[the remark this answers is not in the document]
herjy: Well, that function also matches the physical support of both psfs. I kept the target psf inside this loop just in case bands might have psfs living in different supports. I am not sure the case will arise much, but everything the function does with that regard is to take high resolution psf and trim it to the common support the case may be. Any way, I am pretty sure we will have to review this at some point if we deal differently between SED in various resolutions.

scarlet/interpolation.py
```diff
@@ -295,6 +293,26 @@ def get_separable_kernel(dy, dx, kernel=lanczos, **kwargs):
...
+    ------
+        coord_hr: Coordinates of the high resolution grid
+        coord_lr: Coordinates of the low resolution grid
+        Fm: Sample at positions coord_lr
+    OUTPUTS:
+    -------
+        result:  interpolated  samples at positions coord_hr
+    '''
+    a, b = coord_hr
+    A, B = coord_lr
```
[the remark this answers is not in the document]
herjy: I made all the changes throughout my functions and classes. I leave this open because I am unable to do the tests at the moment.

scarlet/observation.py
```diff
@@ -231,3 +253,146 @@ def get_scene(self, scene):
...
+
+
+                #Computes spatially matching observation and target psfs. The observation psf is also resampled to the scene's resolution
+
+                new_target, observed_psf = resampling.match_psfs(_target, _psf, whr, wlr)
+                #Computes the diff kernel in Fourier
+                target_fft = np.fft.fft2(np.fft.ifftshift(new_target))
+                observed_fft = np.fft.fft2(np.fft.ifftshift(observed_psf))
+                kernel_fft = np.zeros(target_fft.shape)
+                kernel_fft[target_fft != 0] = observed_fft[target_fft != 0] / target_fft[target_fft != 0]
```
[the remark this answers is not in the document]
herjy: THis solved errors when applying the algorithm, big time!

scarlet/observation.py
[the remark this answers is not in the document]
herjy: I originally thought we would need this for representations and figures, but actually, I cannot make a compleing case for it anymore, I a happy to drop it.

scarlet/resampling.py
```diff
@@ -0,0 +1,239 @@
...
+    convolved_fft = model_fft * psf_fft
+    convolved = np.fft.ifft2(convolved_fft)
+    result = np.fft.fftshift(np.real(convolved))
+  #  (bottom, top), (left, right) = padding
+    result = result#[bottom:-top, left:-right]
+    return result
+
+
+
+def conv2D_fft(shape, xm, ym, p, h, scp):
```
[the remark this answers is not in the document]
herjy: This is only used when generating the operator. 
To be honest, my main incentive for using scp here is that it conveniently handles inputs that have different shapes without me having to think of it. :/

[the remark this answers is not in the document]
fred3m: I'm fine with you using `scipy.signal.fftconvolve` here for now, I just think that you should import the module in `conv2D_fft` instead of passing it as a parameter from another method.

scarlet/resampling.py
```diff
@@ -0,0 +1,239 @@
...
+    mat = np.zeros((n1*n2, B.size))
+
+    h = a[1]-a[0]
+    if h == 0:
+        h = b[1]-b[0]
+    assert h !=0
+    import scipy.signal as scp
+    for m in range(np.size(B)):
+            mat[:, m] = conv2D_fft(shape, A[m], B[m], p, h, scp).flatten()*A.size/a.size
+            #mat[:, m] /= np.sum(mat[:,m])
```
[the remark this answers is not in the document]
herjy: removed

scarlet/interpolation.py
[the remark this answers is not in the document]
herjy: And it works

scarlet/interpolation.py
herjy: newaxis trick works

scarlet/observation.py
```diff
@@ -231,3 +253,146 @@ def get_scene(self, scene):
...
+                kernel = np.fft.ifft2(kernel_fft)
+                kernel = np.fft.fftshift(np.real(kernel))
+                diff_psf = kernel/kernel.max()
```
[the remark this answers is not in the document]
herjy: So, I agree with you in principle and I cannot figure out how I came up with the max, but the sum seems to artificially shrink the flux when projecting from high to low resolution.

scarlet/observation.py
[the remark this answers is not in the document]
herjy: Remember, here observed and target psfs are not the same at all! So zero values in target_fft will create infinitely high frequencies. The result, is that without discarding null values, the ringing is made worse. Also, it introduces a  shift. I think I know why, but I'll take a look at that.

scarlet/observation.py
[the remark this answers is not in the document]
pmelchior: please remove then

scarlet/observation.py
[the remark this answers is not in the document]
pmelchior: This is OK if the target_fft has zeros, i.e. no power at some frequencies, which means it's well-resolved. I'd suggest to rewrite as 
```
sel = target_fft != 0
kernel_fft[sel] = observed_fft[sel] / target_fft[sel]
```
to avoid having to do the selection three times.

scarlet/observation.py
```diff
@@ -164,64 +191,59 @@ def weights(self):
...
             The convolved and resampled `model` in the observation frame.
         """
-        if self.wcs is not None:
-            msg = "get_model is currently only supported when the observation frame matches the scene"
-            raise NotImplementedError(msg)
+        if self.structure is not None:
+            assert self.structure.size == model.shape[0]
+            model = model[self.structure==1]
+        if self.psfs is not None:
+            def _convolve_band(model, psf_fft):
```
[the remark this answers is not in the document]
fred3m: This still needs to be addressed. This should be a class method.

scarlet/observation.py
```diff
@@ -231,3 +253,146 @@ def get_scene(self, scene):
...
+        if target_psf is not None:
+            self.target_psf = target_psf
+        else:
+            self.target_psf = target_psf
+        self.structure = structure
+
+
+    def match(self, scene):
+
+        if self.wcs == None:
```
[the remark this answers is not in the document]
fred3m: this was also not fixed

scarlet/observation.py
```diff
@@ -228,11 +228,35 @@ def get_scene(self, scene):
         return self.images
 
 
+class ObservationToResample(Scene):
```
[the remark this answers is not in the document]
herjy: I don't like resampled observation, because it sounds like a model, not an observation. I'll go with LowRes then

scarlet/observation.py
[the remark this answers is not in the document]
herjy: Exactly, well resolved cases should be handled properly. If if not well resolved, it doesn't hurt.

scarlet/observation.py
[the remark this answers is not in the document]
herjy: Done

scarlet/source.py
```diff
@@ -178,7 +178,7 @@ def detection_combination(sed, bg_rms, observations, scene, thresh=1):
     """
 
     for c, obs in enumerate(observations):
-        if type(obs).__name__ is not 'ObservationToResample':
+        if type(obs).__name__ is not 'LowResObservation':
```
pmelchior: That works, but is not normal way of checking. Please use `if isinstance(obs, LowResObservation) is False`, which will need `from .observation import LowResObservation` at the beginning

pmelchior: I also don't think this is what you want to do in the general case of having multiple observations. The logic here is that you want the highest resolution observation to define the morphology, but it may well be the case that none of the observations have the resolution we want to use internally, so they're all LowRes.

pmelchior: Finally, the observation for the detection coadd is identified beforehand with `obj_idx`, why is this for loop even necessary?

docs/resampling.ipynb
```diff
@@ -110,7 +110,7 @@
     "catalog_hsc, bg_rms_hsc = makeCatalog(data_hsc)\n",
```
pmelchior: In `makeCatalog` and elsewhere: why are all those `byteswap().newbyteorder()` necessary?

docs/resampling.ipynb
```diff
@@ -0,0 +1,380 @@
...
+    "print(wcs_hst.wcs.crval)\n",
+    "psf_hst = fits.open('../data/test_resampling/PSF_HST.fits')[0].data\n",
+    "np1,np2 = np.shape(psf_hst)\n",
+    "psf_hst = psf_hst.reshape(1,np1,np2)\n",
+    "\n",
+    "noise = np.concatenate((np.concatenate((data_hst[0,:], data_hst[:,0])),\n",
+    "                                            np.concatenate((data_hst[-1,:], data_hst[:,-1]))))\n",
+    "\n",
+    "n1,n2 = np.shape(data_hst)\n",
+    "data_hst = data_hst.reshape(1, n1, n2).byteswap().newbyteorder()*np.max(data_hsc)/np.max(data_hst)\n",
```
pmelchior: Why are the values in `data_hst` and `data_hsc` normalized so that they have the same maximum?

docs/resampling.ipynb
```diff
@@ -0,0 +1,380 @@
...
+    "Each source is a list of fundamental `scarlet.Component` instances and must be based on `scarlet.Source` or a derived class, in this case `ExtendedSource`, which enforces that the source is monotonic and symmetric."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "obs_hst = scarlet.Observation(data_hst, wcs = wcs_hst, psfs = None, structure = np.array([0,0,0,1]))\n",
```
pmelchior: @fred3m Not to be addressed in this PR, but something we should correct: `psfs=None` is the only way to say: don't convolve the model. What we should rather say `if psfs is None or psfs = blend.psfs: don't convolve`

Then we could properly say that this Observation actually has a PSF, it just is the same one as our target.

docs/resampling.ipynb
```diff
@@ -0,0 +1,380 @@
...
+    "\n",
+    "# In[5]:\n",
+    "\n",
+    "sources = [scarlet.CombinedExtendedSource((ra[i], dec[i]), scene, obs, bg_rms, symmetric = 0, monotonic = 1) for i in range(ra.size)]\n",
+    "#make condition on combination and observations\n",
+    "#Do the get-match here in the initialisation. Linear solution for the SEDs.\n",
+    "\n",
+    "rgb = np.zeros((4,n1,n2))\n",
+    "\n",
+    "for src in sources:\n",
```
pmelchior: Here we should have a custom `filter_weights` argument for `img_to_rgb` so that HST and HSC filters looks reasonable together. Not critical for this PR

docs/resampling.ipynb
fred3m: I don't see what you gain by that. Not specifying a PSF is just another way of saying we are not doing a convolution, so the result is exactly the same. I think it is simpler to say that if the PSF of the scene and the model are the same then initialize the `Observation` with `psfs=None`.

docs/resampling.ipynb
fred3m: See https://sep.readthedocs.io/en/v1.0.x/tutorial.html#Finally-a-brief-word-on-byte-order. This is an issue with the way that astropy handles endianness: http://docs.astropy.org/en/stable/known_issues.html#bug-with-unicode-endianness-in-io-fits-for-big-endian-processors

scarlet/observation.py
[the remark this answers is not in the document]
fred3m: I think that we need to understand this better. It doesn't need to hold up this PR but we should open an issue to investigate this. Ideally we'd like to be able to make measurements on the pre-convolved models but a scaling during convolution will prevent us from doing that.

docs/resampling.ipynb
pmelchior: We'd gain consistency. If we want to match PSFs (from model to observation), we give the PSF argument, if not, we omit it. So far so, good.

What happens if I want to change the PSF of `Blend`, i.e. of our internal model? I now need to remember to specify the PSFs for all observations. If I forget, the result will be incorrect. Therefore the observation should be fully specified. If an Observation now has the same PSF as `Blend`, it won't convolve. It's a very simple test that adds consistency to the declaration of observations:
```
obs_hst = scarlet.Observation(data_hst, wcs = wcs_hst, psfs = psfs_hst)
obs_hsc = scarlet.LowResObservation(data_hsc, wcs = wcs_hsc, psfs = psfs_hsc)
```

fred3m: I agree but I think that this only works if we create a `PSF` class. Otherwise it has to check the PSFs element by element, which is ~43^2 pixels for an HSC PSF and much larger when it is upsampled to HST resolution.

scarlet/source.py
herjy: That will be because the whole point of having obs_idx was lost on me. Actually, I can get rid of the whole function and use build_detection_coadd here

docs/resampling.ipynb
herjy: idk
Something used to complain about the type, but now it seems fine.

docs/resampling.ipynb
herjy: Strictly speaking, we don't need this. But let's say that one wants to make an rgb image that uses images from different datasets in the data frame, this avoids to have one band overpowering all the others.

docs/resampling.ipynb
pmelchior: I'm pretty sure it would just do a reference check because both point to the same PSF structure

docs/resampling.ipynb
pmelchior: that's what the new `filter_weights` concept for RGB images does. In fairness, we just merged it into master yesterday ...

As a general rule, observed data should receive very little modification, and only if it's scientifically motivated

PR #88, opened 2019-06-06
proper multi-band images and RGB scaling; closes #85

### LINKED ISSUE #85: more than 3 filters for RGB images

pmelchior:


scarlet/display.py
```diff
@@ -1,224 +1,268 @@
...
-    sample_slice = slice(int(np.floor((0.5-radius)*n_pix)), int(np.ceil((0.5+radius)*n_pix)))
-    result = np.polyfit(idx[sample_slice], sorted_img[sample_slice], 1)
-    slope, intercept = result
+    maxRGB = 255
+    return np.dstack((rgb * maxRGB)).astype(np.uint8)
 
-    z1 = max(median - center * (slope/contrast), np.min(img))
-    z2 = min(median + (n_pix-center-1) * (slope/contrast), np.max(img))
-    return z1, z2
+def rgb(img, filter_weights=None, fill_value=0, norm=None):
```
fred3m: I think this should be called `img_to_rgb`. You use the parameter `rgb` in some of your other functions and this naming convention makes it clear what this function does.

scarlet/display.py
```diff
@@ -1,224 +1,268 @@
...
+    norm: `scarlet.display.Norm`, default `None`
+        Norm to use for mapping in the allowed range [0..255]. If `norm=None`,
+        `scarlet.display.LinearPercentileNorm` will be used.
+
+    Returns
+    -------
+    rgb: numpy array with dimensions (3, height, width) and dtype uint8
+    """
+    RGB = img_to_channel(img)
+    if norm is None:
```
fred3m: Why not just make the default value for `norm=LinearPercentileNorm`? The only time you should need to use `None` in the function signature is if the default value is an iterable or has a more complicated default value (like `filter_weights`).

pmelchior: That's because you don't need the class of the norm, but an actual instance. The use case is if you define a survey-wide norm, linear or asinh, which you then need to pass to this method. As is, the method instantiates the norm if it is `None`, but it could not or should not do that all the time.

scarlet/display.py
```diff
@@ -1,224 +1,268 @@
...
+        Parameters
+        ----------
+        img: array_like
+            Image to normalize
+
+        Returns
+        -------
+        normalized img
+        """
+        img -= self.vmin
```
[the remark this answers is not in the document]
pmelchior: That's better, yes. It was OK since this method is general, but it is *meant* to be run on the RGB channel values, not the original images. See `img_to_rgb`.

pmelchior:
I will merge the PR now, but I wanted to flag that the image scaling will be somewhat different now. For the docs update we should check that the images come out OK.

scarlet/display.py
fred3m: You're right, I didn't catch that.

PR #92, opened 2019-06-12
tickets/DM-20030

fred3m:
A few bug fixes and a lot of updates to all of the docs to be compatible with scarlet v0.5. I also changed from Read-The-Docs to github pages for the documentation, which gives us more control over the build environment and is more stable. I also added a script to build and update the docs and updated the `install` page to describe the process.


pmelchior:
This looks very good. A few things I would change in the docs:
* in the left table of contents column, I'd remove "Building the Docs". Most people won't do it and it's the prominent 2nd item in the list.
* in the quickstart [visualization section](https://fred3m.github.io/scarlet/quickstart.html#View-the-source-models), the stretch is set to the model max. That's not a good idea because it changes the amplitude between the different components, and for faint sources it looks awful. [Plus: it's clear from model1 why we need symmetry back]
* I'm not sure anyone needs the "Diagnostics" section.

fred3m:
Ok, I fixed the table of contents. The visualization is tricky, the faint objects are *really* faint, so when I try to use the same scaling as the other sources they were pretty much invisible. I played around for half a day with different stretches and Q values and determined that there is no ideal way to display those sources. So I chose a stretch that works best for most of the sources and I think the key takeaway is that the other sources are so faint that we can't really distinguish them from the noise in the original image. So if you feel strongly about changing the scaling I can... but it might be better to just not display the faint sources.

I also removed the diagnostics section. Originally I wanted this to be a place to illustrate how different algorithms affect the results but I think that there might be a more appropriate place to put them, like in the stack club repository.

pmelchior:
The TOC is still buggy. When you go to e.g. Tutorial or User Guide, you can still "Building the Docs" on the left.

As for the RGB images: We should show all of the model in the same stretch. I can see all of them in the second row, so that's a useful stretch. [Btw, the difficulty of setting the stretch is my main objection against this parameterization.] If we can't see the faint but detected sources in the RGB, we need to tinker with the settings, but only once.

+1 on the diagonstics

fred3m:
Looks like sphinx tried a little too hard to be efficient and only rebuilt the pages that changed. I'm surprised that they didn't realize the TOC needed to be rebuilt, so I wiped the build and rebuilt from scratch, which fixed that issue (you might have to clear your browser cache or use a private window, since browsers also try to be too clever sometimes).

I used the largest stretch that will show all of the sources. It amplifies the background significantly in the data image, but if you insist on having all of them using the same stretch then this is really the best it can do.

pmelchior:
Good, thanks. Waiting for the checks to complete, then I'll merge

Issue #95, opened 2019-06-20
Check n-dim FFTs for convolution operations


pmelchior:
Currently we do B two-dimensional FFTs when convolving the model in `Observation.get_loss`, one for each band. However, according to the [numpy docs](https://docs.scipy.org/doc/numpy/reference/generated/numpy.fft.rfftn.html#numpy.fft.rfftn) one can perform a N-dim FFT along specified axes, which would allow us to transform the entire model data cube and PSF data cube in one go, multiply, and then reverse the transform again in one go. The `axes` and `s` parameters should be useful.

However, we don't use the default numpy, we use `autograd.numpy`, which has that function implemented, but it may not be as efficient as the default.

So, we need to test and find out with a profiler.

PR #99, opened 2019-06-22
tickets/DM-20128: Create tests

fred3m:
This ticket creates tests for the bulk of scarlet classes and functions.


fred3m:
This ticket contains tests for the bulk of scarlet. I would still like to come up with a good test for initializing `ExtendedSource` and ensuring that the `autograd` functions are working as they should, but this dramatically increases our test coverage.

I decided not to include tests for building the docs since that looks a bit complicated and we have more pressing changes to implement. Once those are pushed to master we can revisit building the docs on Travis and adding a flake8 test to ensure a consistent coding style.

pmelchior:
Before looking at the code, that all checks have failed on this branch is not a good sign...

docs/install.rst
```diff
@@ -46,6 +46,20 @@ If you already have a local version of Eigen_ and don't want to download the hea
     `build_ext` does not accept relative paths, so `<full path to your Eigen header files>`
     must be a full path.
 
+
+For OS X Users
+--------------
+
+For some reason on OS X the system attempts to use the debprecated stdlc++ library instead of the
```
pmelchior: deprecated

scarlet/observation.py
```diff
@@ -105,74 +124,64 @@ class Observation(Scene):
         prevent artifacts due to the FFT.
     """
 
-    def __init__(self, images, psfs=None, weights=None, wcs=None, filtercurve=None, padding=3, structure=None):
+    def __init__(self, images, psfs=None, weights=None, wcs=None, filtercurve=None, structure=None,
+                 padding=10):
         super().__init__(images.shape, wcs=wcs, psfs=psfs, filtercurve=filtercurve)
 
         self.images = np.array(images)
```
pmelchior: why are images, weights, PSFs copied? They should never be altered, so changes should not be necessary.

scarlet/observation.py
```diff
@@ -105,74 +124,64 @@ class Observation(Scene):
...
-                # we may be able to come up with a method to normalize in the Fourier Transform
-                # and avoid this step.
-                kernel = np.fft.ifft2(kernel_fft)
-                kernel = np.fft.fftshift(np.real(kernel))
-                kernel /= kernel.sum()
-                # Store the Fourier transform of the matching kernel
-                new_kernel_fft.append(np.fft.fft2(np.fft.ifftshift(kernel)))
+            kernels = []
+            for psf in self.psfs:
+                _psf_fft = np.fft.rfftn(psf, fftpack_shape)
```
pmelchior: could we do the FFT with all bands at once and then split into the bands if needed?

pmelchior: In fact, it looks like the entire `for` block could be done with all bands at once, which would also avoid the extra copies in lines 174-175

pmelchior: This is not required work for this PR since it's #95

scarlet/source.py
```diff
@@ -35,59 +35,22 @@ def get_pixel_sed(sky_coord, observations):
...
 
     sed = []
-    band = 0
     for obs in observations:
         pixel = obs.get_pixel(sky_coord)
+        _sed = obs.images[:, pixel[0], pixel[1]].copy()
+        if obs.psfs is not None:
+            # Account for the PSF in the intensity
+            _sed /= obs.psfs.max(axis=(1, 2))
+        sed = np.concatenate((sed, _sed))
```
pmelchior: this is an implicit ordering, required because #100 isn't yet implemented

pmelchior: not to be change here, only to flag the issue

scarlet/observation.py
pmelchior: Let's make this a full code review: #101 

Not to be changed here

fred3m:
> Before looking at the code, that all checks have failed on this branch is not a good sign...

This is why I said that proxmin had to be merged first. All tests that depended on the correct proxmin behavior (eg. inline updates) fail. Now that proxmin has been merged I changed the version of proxmin required to install scarlet so everything should build, since it builds on my local computer.

scarlet/observation.py
fred3m: Wrapping a sequence with `np.array` doesn't copy the object, it just creates a view to the contents of the sequence with the numpy array methods. So if someone passed a list to `images`, for example, then calling `B, Ny, Nx = images.shape` would raise an Exception if we didn't use the `np.array` wrapper. Internally any numpy function that you call (eg. `np.sum`) automatically does this at very little cost, so I don't think that it should be an issue. Numpy only copies if you use the `copy` method.

scarlet/observation.py
fred3m: I agree that this *should* be possible.

PR #104, opened 2019-06-24
Better naming convention; consistency in multi-observation cases

pmelchior:
Fixes four issues referenced in the commit logs


scarlet/observation.py
```diff
@@ -22,54 +22,54 @@ def _centered(arr, newshape):
...
-            # Make sure that psf is always 3D
-            if len(psfs.shape) == 2:
-                psfs = psfs[None]
-            psfs = psfs / psfs.sum(axis=(1, 2))[:, None, None]
+        if psfs is None:
+            logger.warning('No PSFs specified. Possible, but dangerous!')
+        else:
+            assert len(psfs) == 1 or len(psfs) == shape[0]
+            if not np.allclose(psfs.sum(axis=(1, 2)), 1):
+                logger.warning('PSFs not normalized!')
```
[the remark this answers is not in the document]
fred3m: Given our conversation on slack this could be changed to 
```
if not np.allclose(psfs.sum(axis=(1, 2)), 1):
    logger.warning('PSFs not normalized!')
    psfs[:] = psfs / psfs.sum(axis=(1, 2))[:, None, None]
```
which would make sure that the PSF is properly normalized without creating a new instance.

fred3m:
My review was just for the code changes. I'll review the docs and test changes tomorrow.

scarlet/observation.py
```diff
@@ -22,54 +22,54 @@ def _centered(arr, newshape):
...
+
         self._psfs = psfs
-        assert filtercurve is None or shape[0] == len(filtercurve)
-        self.filtercurve = filtercurve
+
+        assert bands is None or len(bands) == shape[0]
```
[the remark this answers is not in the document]
herjy: I'm with Fred on that one, plus bands also refers to an image observed in a given filter. While there is no ambiguity in the code properly speaking, I find it semantically dodgy.

scarlet/observation.py
```diff
@@ -101,156 +99,171 @@ def get_pixel(self, sky_coord):
...
-    def match(self, scene):
-        """Match the psf in each observed band to the target PSF
+    def match(self, model_frame):
+        """Match the frame of `Blend` to the frame of this observation.
+
+        The method sets up the mappings in spectral and spatial coordinates,
+        which includes a spatial selection, computing PSF difference kernels
+        and filter transformations.
+
+        Arguments
```
fred3m: This needs to be `Parameters`.

scarlet/observation.py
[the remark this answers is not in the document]
pmelchior: +1, but then we need to change the warning to say that we just did normalize the PSF

scarlet/observation.py
```diff
@@ -259,133 +272,105 @@ def __init__(self, images, wcs=None, psfs=None, weights=None, filtercurve=None,
...
+        resconv_op = []
+        for _psf in self.frame.psfs:
+            # Computes spatially matching observation and target psfs. The observation psf is also resampled to the model frame resolution
+            new_target, observed_psf = resampling.match_psfs(_target, _psf, whr, wlr)
+            # Computes the diff kernel in Fourier
+            target_fft = np.fft.fft2(np.fft.ifftshift(new_target))
+            observed_fft = np.fft.fft2(np.fft.ifftshift(observed_psf))
+            kernel_fft = np.zeros(target_fft.shape)
+            sel = target_fft != 0
+            kernel_fft[sel] = observed_fft[sel] / target_fft[sel]
```
fred3m: This line generates a `ComplexWarning`:
```
/Users/fred/projects/scarlet/scarlet/observation.py:308: ComplexWarning: Casting complex values to real discards the imaginary part
  kernel_fft[sel] = observed_fft[sel] / target_fft[sel]
```
I think this is because `kernel_fft` is not initialized as a complex array, which it needs to be because `observed_fft` and `target_fft` are both complex. This might solve the color issue that you described in slack.

scarlet/observation.py
```diff
@@ -22,54 +22,54 @@ def _centered(arr, newshape):
...
-        if psfs is not None:
-            psfs = np.array(psfs)
-            # Make sure that psf is always 3D
-            if len(psfs.shape) == 2:
-                psfs = psfs[None]
-            psfs = psfs / psfs.sum(axis=(1, 2))[:, None, None]
+        if psfs is None:
+            logger.warning('No PSFs specified. Possible, but dangerous!')
+        else:
+            assert len(psfs) == 1 or len(psfs) == shape[0]
```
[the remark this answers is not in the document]
pmelchior: It's my crummy way to solve the problem of having one PSF for `Blend` and multiple ones for `Observation`. It would be nice to automate turning a 2D PSF image into a 3D, but you can't alter the `psfs` with a view (e.g. `psfs =  psfs.reshape(1,Ny,Nx)`) without breaking the comparison.

docs/tutorials/point_source.ipynb
```diff
@@ -162,11 +163,11 @@
...
    "source": [
     "# Initialize the Blend object, which later fits the model\n",
-    "blend = scarlet.Blend(scene, sources, observation)\n",
+    "blend = scarlet.Blend(frame, sources, observation)\n",
     "\n",
     "# Display the initial model\n",
-    "_model = blend.get_model()\n",
-    "model = observation.get_model(_model)\n",
+    "model = blend.get_model()\n",
+    "model_ = observation.render(model)\n",
```
fred3m: Something isn't right. It doesn't appear to be properly matching the observation PSFs, causing the output to be in the model frame, not the observation frame. Compare the results of this notebook with https://fred3m.github.io/scarlet/tutorials/point_source.html#Fit-the-model-and-display-the-results

scarlet/observation.py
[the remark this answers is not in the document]
pmelchior: Problem is that `filter` is a special term for python. So writing something like `for filter in filtercurves` is problematic, too. I'm actually not convinced that the naming is dodgy. *Filter* refers to a color piece of glass, which spectrographs don't use. *Band* is the outcome of it. I think this is reasonable.

docs/tutorials/point_source.ipynb
fred3m: I found the bug. In both this cell and the next one you use `model`, not `model_`.

tests/test_observation.py
```diff
@@ -30,82 +30,85 @@ def get_psfs(self, shape, sigmas):
...
         psfs, truth = self.get_psfs(shape, [2.1, 1.1, 3.5])
+        psfs /= psfs.sum(axis=(1,2))[:,None,None]
 
-        scene = scarlet.Scene(psfs.shape, psfs=target_psf)
+        frame = scarlet.Frame(psfs.shape, psfs=target_psf)
         observation = scarlet.Observation(psfs, psfs)
-        observation.match(scene)
-        result = observation.get_model(np.array([target_psf]*len(psfs)))
+        observation.match(frame)
+        result = observation.render(np.array([target_psf]*len(psfs)))
```
fred3m: Ok, I figured out why this test is failing. You added line 95, which changes the target PSF shape to go from (Ny, Nx) to (1, Ny, Nx), so when they are compared in the assert statement they have different lengths. This is because one of the things this was testing is that `Scene` assumes that a 2D input PSF is a single band PSF and adds the extra dimension, which the new version of the code no longer does. Rather than make a change myself that might cause a conflict in your branch, I'll give you the updated code below:

```
def test_psf_match(self):
    shape = (43, 43)
    target_psf = self.get_psfs(shape, [.9])[1]
    psfs, truth = self.get_psfs(shape, [2.1, 1.1, 3.5])
    psfs /= psfs.sum(axis=(1,2))[:,None,None]

    frame = scarlet.Frame(psfs.shape, psfs=target_psf)
    observation = scarlet.Observation(psfs, psfs)
    observation.match(frame)
    result = observation.render(np.array([target_psf[0]]*len(psfs)))

    assert_almost_equal(result, truth)
```

tests/test_blend.py
```diff
@@ -37,6 +37,8 @@ def init_data(shape, coords, amplitudes=None, convolve=True):
 
         psfs = np.array([scarlet.psf.generate_psf_image(scarlet.psf.gaussian, psf_shape, psf_center,
                                                         amplitude=1, sigma=1+.2*b) for b in range(B)])
+        psfs /= psfs.sum(axis=(1,2))[:,None,None]
```
fred3m: This is the line that breaks the other two tests. The problem is that `images` is calculated using the non-normalized PSFs, since we want it to have the same amplitude as the SEDs calculated above (so the psf was a gaussian with *amplitude* unity). So if you just move this line below line 44 and normalize after `images` is created then the tests pass.

docs/tutorials/point_source.ipynb
fred3m: Actually, it's probably better to use
```
_model = blend.get_model()
model = observation.render(_model)
```
so that you only have to make one change. Otherwise you not only have to change the way you create the RGB images, you also have to modify the line where you calculate the residual.

docs/user_docs.ipynb
```diff
@@ -159,8 +162,17 @@
...
-    "scene = scarlet.Scene(images.shape, psfs=target_psf)"
+    "frame = scarlet.Frame(images.shape, psfs=model_psf, bands=filters)\n",
+    "observation = scarlet.Observation(images, psfs=psfs, bands=filters).match(frame)"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {},
+   "source": [
+    ".. note::\n",
```
fred3m: You cannot insert an RST `note` or `warning` in a cell, because ipython uses markdown, not restructured text. This is one of the PITA parts of writing documents using ipython. So you have to put the note in its own cell, choose cell type = raw, view the type of raw input cell, and select rst from the dropdown, all in ipython, otherwise the note isn't rendered properly (see the built docs).

docs/user_docs.ipynb
```diff
@@ -796,13 +799,11 @@
...
-    "sources = [scarlet.ExtendedSource((src['y'],src['x']), scene, observation, bg_rms) for src in catalog]\n",
-    "blend = scarlet.Blend(scene, sources, observation)\n",
-    "blend.fit(20)\n",
+    "blend.fit(10)\n",
     "print(blend.it) \n",
-    "blend.fit(20)\n",
+    "blend.fit(10)\n",
     "print(blend.it)\n",
-    "blend.fit(20)\n",
+    "blend.fit(10)\n",
```
fred3m: Good catch. You should also change the text that follows from "...60 iterations" to "...30 iterations".

fred3m:
Ok, I'm finished with my review now.

scarlet/observation.py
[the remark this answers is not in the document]
pmelchior: As we agreed offline, I will rename `bands` to `channels` consistently throughout the code

scarlet/observation.py
pmelchior: Problem for @herjy to fix in the new FFT code

herjy:
Sorry, I only did a shallow pass at the code. I don't have much to add other than what Fred said. Do you need me to give a more indepth look? 
The thing is that I will do a PR just after we pull this one and will update according to your your changes here, so this will give me a second chance at commenting your changes.

scarlet/observation.py
herjy: It is already fixed on my branch

scarlet/observation.py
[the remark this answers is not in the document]
pmelchior: Made more verbose

scarlet/observation.py
```diff
@@ -101,156 +99,171 @@ def get_pixel(self, sky_coord):
...
+            The frame of `Blend` to match
+
+        Returns
+        -------
+        None
         """
-        if self.psfs is not None:
+
+        #  bands of model that are represented in this observation
+        self.band_slice = slice(None)
```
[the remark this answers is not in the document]
pmelchior: no, but `_bands_slice` is better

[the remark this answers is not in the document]
pmelchior: that being said, we have a lot of other attributes of `Observation` that are not important for a user to know about. I'll hide them too

docs/tutorials/point_source.ipynb
pmelchior: No problem, but I prefer it this way: `model` is direct from `Blend`, `model_` is after it gets mangled by `render`

docs/user_docs.ipynb
pmelchior: Moved to ordinary markdown

fred3m:
Ok, this looks good to me now. All of the tests pass and I built the docs locally and can push them once you merge the branch.

scarlet/component.py
```diff
@@ -267,22 +274,13 @@ def K(self):
...
-
-    @property
-    def Nx(self):
-        """Number of pixels in the y-direction
-        """
-        return np.max([comp.Nx for comp in self.components])
+        try:
+            return self._frame
+        except AttributeError:
+            return self.components[0].frame
```
fred3m: If the `ComponentTree` does not have a `_frame` attribute then I would actually institute a check here to ensure that all of the components have the same frame , then set `_frame` to that frame.

scarlet/source.py
```diff
@@ -238,50 +222,45 @@ def init_multicomponent_source(sky_coord, frame, observations, bg_rms, flux_perc
...
-        try:
-            iter(observations)
-        except TypeError:
-            observations = [observations]
+        """Extended source intialized with a single pixel
+
+        Point sources are initialized with the SED of the center pixel,
+        and the morphology of a single pixel (the center) turned on.
+        While the source can have any `constraints`, the default constraints are
+        symmetry and monotonicity.
```
fred3m: This should be in the docstring for the class. The only thing in the constructor docstring should be the parameters.

tests/test_blend.py
```diff
@@ -79,26 +72,21 @@ def test_init(self):
...
+        frame = scarlet.Frame(images.shape, psfs=target_psf[None])
+        obs = scarlet.Observation(images, psfs=psfs).match(frame)
+        sources = [scarlet.PointSource(frame, coord, observation) for coord in coords]
 
         with pytest.raises(AttributeError):
-            obs1.psfs_fft
-        with pytest.raises(AttributeError):
-            obs2.psfs_fft
-        blend = scarlet.Blend(frame, sources, observations)
+            obs.psfs_fft
```
fred3m: You can remove this test (and the pytest import at the top of the file). This test was in place when we ere using `Blend` to do the PSF matching and checked to make sure that the PSFs had't been matched yet. We don't use `psfs_fft` anymore, so this is a useless test now.

tests/test_blend.py
```diff
@@ -79,26 +72,21 @@ def test_init(self):
         assert blend.frame.psfs == frame.psfs
 
         # Test init with psfs
-        frame = scarlet.Frame(images.shape, psfs=target_psf[None], channels=channels)
-        obs1 = scarlet.Observation(images1, psfs=psfs1, channels=channels1).match(frame)
-        obs2 = scarlet.Observation(images2, psfs=psfs2, channels=channels2).match(frame)
-        observations = [obs1, obs2]
-        sources = [scarlet.PointSource(coord, frame, observations) for coord in coords]
+        frame = scarlet.Frame(images.shape, psfs=target_psf[None])
+        obs = scarlet.Observation(images, psfs=psfs).match(frame)
```
fred3m: The test is failing because you are using inconsistent names for the rest of this test. Here you define `obs`, but in some lines of code below you use `obs` and in others you use `observation`, which does not have a PSF and thus has different scaling.

scarlet/component.py
pmelchior: True, but it's a somewhat contrived scenario. The only reason this isn't merely a `return self._frame` is that we would initialize `Blend.frame` as a `ComponentTree` from the sources alone. `Blend` wouldn't need the `frame` argument at all. This is a change of dependency logic, but I'm open for discussion.

We currently say `frame` is the frame for Blend as a model. So, it should better have `frame` as argument.

But we can also say, `frame` is the frame for the sources/components as the fundamental units of the model, and `Blend` is just their collection. Then `Blend` doesn't need `frame`.

fred3m: I disagree. I think that all of the components need to have the same `Frame`, otherwise `ComponentTree.get_model` doesn't work. It also breaks the interchangeability of `Component` and `ComponentTree`, which depends on `ComponentTree` having all of the same attributes and methods as `Component` (plus additional methods needed for the parent) in order to allow sources/components to be hierarchical. 

So the tree seems to be the natural place to define the `Frame` for all of its components and check that they all agree. Really this should probably be done in `ComponentTree.__init__` and `Component.frame` should just return `self._frame`. We should also check that two components/trees have equal frames in `__iadd__`, and implement a few new tests to make sure that initialization and addition fail if they don't.

tests/test_blend.py
pmelchior: Even with fresh eyes this morning, I didn't see the mistake, but sure enough ...

scarlet/component.py
fred3m: (but the tests can be done in a separate PR)

tests/test_blend.py
fred3m: It does help to have written all of the tests so far... :)

scarlet/component.py
pmelchior: The consistency check is not a problem, and a good idea.

But the sources need `frame` to get initialized, the question is does `Blend`? If we decide, we treat it in the tree, which makes sense to me, too, `Blend` doesn't need it, and we should remove it now while we're making API changes.

PR #105, opened 2019-06-27
RandomSource; closes #98

### LINKED ISSUE #98: Create `RandomSource`
To test the importance of custom initialization, it is useful to have a source that has a randim uniform morphology, with a SED read of by projecting into the observation frames.

The first part is trivial, the second requires functionality that we might want to have anyways.

pmelchior:
minimum implementation of a randomly initialized source. can take SED by fitting to an observation if provided


scarlet/source.py
```diff
@@ -227,6 +227,36 @@ def init_multicomponent_source(sky_coord, frame, observation, bg_rms, flux_perce
...
+        frame: `~scarlet.Frame`
+            The frame of the model
+        observation: list of `~scarlet.Observation`
+            Observation to initialize the SED of the source
+        component_kwargs: dict
+            Keyword arguments to pass to the component initialization.
+        """
+        C, Ny, Nx = frame.shape
+        morph = np.random.rand(Ny, Nx)
+
```
herjy: I usually like to scale random optimisation by the amplitude of the signal, that usually makes convergence slightly faster, but that's more of a familly recipe.

scarlet/source.py
```diff
@@ -227,6 +227,36 @@ def init_multicomponent_source(sky_coord, frame, observation, bg_rms, flux_perce
...
+        if observation is None:
+            sed = np.random.rand(C)
```
herjy: I'm not sure we want to have both seds and morphologies initialised at random. Also, not providing an observation would not make much sense, would it?

scarlet/source.py
pmelchior: Amplitudes are in the SED for us

scarlet/source.py
pmelchior: True, this is why you *can* provide an observation, but you don't have to. Double random initialization is not uncommon for matrix factorization, but normally one would scale to the amplitude in each channel, which is what this code will do if given an observation.

In some sense, this part demonstrates that a source can formally exist without an observation to initialize it.

pmelchior:
@herjy if you're ok with it, do the honors and merge the PR. Delete the branch after it's merged.

scarlet/source.py
herjy: I just don't see a use case for no observation initialisation. 

If no observation is given, then nothing is scaled (to nothing because there is nothing to scale to, makes sense...). That's about the worse case scenario for random initialisation. This is actually great for studying random initialisation, but not for production, we agree?

PR #107, opened 2019-07-01
tickets/DM-20431: Implement Nate Lust's symmetry algorithm

fred3m:
Implement symmetry in kspace and fractional pixel shifts
calculated by using the morphology centroid weighted by
the frame psf.

### DIFF
--- a/scarlet/operator.py
+++ b/scarlet/operator.py
@@ -1,5 +1,6 @@
 from functools import partial
 
+from scipy import fftpack
 import autograd.numpy as np
 from proxmin.operators import prox_unity_plus
 from proxmin.utils import MatrixAdapter
@@ -238,7 +239,7 @@ def prox_sdss_symmetry(X, step):
 
 def prox_soft_symmetry(X, step, strength=1):
     """Soft version of symmetry
-    Using a `sigma` that varies from 0 to 1,
+    Using a `strength` that varies from 0 to 1,
     with 0 meaning no symmetry enforced at all and
     1  being completely symmetric, the user can customize
     the level of symmetry required for a component
@@ -248,7 +249,56 @@ def prox_soft_symmetry(X, step, strength=1):
     return X
 
 
-def prox_uncentered_symmetry(X, step, center=None, strength=.5, use_soft=True, fill=None):
+def prox_kspace_symmetry(X, step, shift=None, padding=10):
+    """Symmetry in Fourier Space
+
+    This algorithm by Nate Lust uses the fact that throwing
+    away the imaginary part in Fourier space leaves a symmetric
+    soution in real space. So `X` is transformed to Fourier space,
+    shifted by the fractional amount `shift=(dy, dx)`,
+    the imaginary part is discarded, shited back to its original position,
+    then transformed back to real space.
+    """
+    # Record the morph shape
+    shape = X.shape
+    dy, dx = shift
+    padding = np.max(X.shape) + padding // 2
+    edges = ((padding, padding), (padding, padding))
+    corner = (padding, padding)
+    zeroMask = X <= 0
+    X[zeroMask] = 0
+    X = np.pad(X, edges, 'constant')
+
+    freq_x = np.fft.fftfreq(X.shape[1])
+    freq_y = np.fft.fftfreq(X.shape[0])
+
+    # Transform to k space
+    X_fft = np.fft.fftn(np.fft.ifftshift(X))
+
+    # Shift the signal to recenter it, negative because math is opposite from
+    # pixel direction
+    shifter = np.outer(np.exp(-1j*2*np.pi*freq_y*-(dy)),
+                       np.exp(-1j*2*np.pi*freq_x*-(dx)))
+    inv_shifter = np.outer(np.exp(-1j*2*np.pi*freq_y*(dy)),
+                           np.exp(-1j*2*np.pi*freq_x*(dx)))
+    result_fft = X_fft*shifter
+
+    # symmeterize
+    result_fft = result_fft.real
+
+    # Shift back
+    result_fft = result_fft*inv_shifter
+
+    # Transform to real space
+    result = np.fft.fftshift(np.fft.ifftn(result_fft))
+    # Return the unpadded transform
+    result = np.real(result[corner[0]:corner[0]+shape[0], corner[1]:corner[1]+shape[1]])
+    result[zeroMask] = 0
+    assert result.shape == shape
+    return result
+
+
+def prox_uncentered_symmetry(X, step, center=None, algorithm="kspace", fill=None, shift=None, strength=.5):
     """Symmetry with off-center peak
 
     Symmetrize X for all pixels with a symmetric partner.
@@ -259,24 +309,55 @@ def prox_uncentered_symmetry(X, step, center=None, strength=.5, use_soft=True, f
         The parameter to update.
     step: `int`
         Step size of the gradient step.
-    center: tuple
+    center: tuple of `int`
         The center pixel coordinates to apply the symmetry operator.
+    algorithm: `string`
+        The algorithm to use for symmetry.
+        * If `algorithm = "kspace" then `X` is shifted by `shift` and
+          symmetry is performed in kspace. This is the only ymmetry algorithm
+          in scarlet that works for fractional pixel shifts.
+        * If `algorithm = "sdss" then the SDSS symmetry is used,
+          namely the source is made symmetric around the `center` pixel
+          by taking the minimum of each pixel and its symmetric partner.
+          This is the algorithm used when initializing an `ExtendedSource`
+          because it keeps the morphologies small, but during optimization
+          the penalty is much stronger than the gradient
+          and often leads to vanishing sources.
+        * If `algorithm = "soft" then soft symmetry is used,
+          meaning `X` will be allowed to differ from symmetry by the fraction
+          `strength` from a perfectly symmetric solution. It is advised against
+          using this algorithm because it does not work in general for sources
+          shifted by a fractional amount, however it is used internally if
+          a source is centered perfectly on a pixel.
+    fill: `float`
+        The value to fill the region that cannot be made symmetric.
+        When `fill` is `None` then the region of `X` that is not symmetric
+        is not constrained.
     strength: `float`
         The amount that symmetry is enforced. If `strength=0` then no
         symmetry is enforced, while `strength=1` enforces strict symmetry
         (ie. the mean of the two symmetric pixels is used for both of them).
-    use_soft: `bool`
-        Whether to use the strict SDSS symmetry `use_soft=False` or
-        the soft proximal operator (`use_soft=True`).
+        This parameter is only used when `algorithm = "soft"`.
 
     Returns
     -------
     result: `function`
         The update function based on the specified parameters.
     """
-    if use_soft:
+    if algorithm == "kspace" and (shift is None or np.all(shift == 0)):
+        algorithm = "soft"
+        strength = 1
+    if algorithm == "kspace":
+        return uncentered_operator(X, prox_kspace_symmetry, center, shift=shift, step=step, fill=fill)
+    if algorithm == "sdss":
+        return uncentered_operator(X, prox_sdss_symmetry, center, step=step, fill=fill)
+    if algorithm == "soft" or algorithm == "kspace" and shift is None:
+        # If there is no shift then the symmetry is exact and we can just use
+        # the soft symmetry algorithm
         return uncentered_operator(X, prox_soft_symmetry, center, step=step, strength=strength, fill=fill)
-    return uncentered_operator(X, prox_sdss_symmetry, center, step=step, fill=fill)
+
+    msg = "algorithm must be one of 'soft', 'sdss', 'kspace', recieved '{0}''"
+    raise ValueError(msg.format(algorithm))
 
 
 def proj(A, B):
--- a/scarlet/operator.py
+++ b/scarlet/operator.py
@@ -1,6 +1,5 @@
 from functools import partial
 
-from scipy import fftpack
 import autograd.numpy as np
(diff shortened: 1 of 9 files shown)


scarlet/operator.py
```diff
@@ -259,24 +309,55 @@ def prox_uncentered_symmetry(X, step, center=None, strength=.5, use_soft=True, f
         The parameter to update.
     step: `int`
         Step size of the gradient step.
-    center: tuple
+    center: tuple of `int`
         The center pixel coordinates to apply the symmetry operator.
+    algorithm: `string`
+        The algorithm to use for symmetry.
+        * If `algorithm = "kspace" then `X` is shifted by `shift` and
+          symmetry is performed in kspace. This is the only ymmetry algorithm
```
herjy: typo on `symmetry`

scarlet/operator.py
```diff
@@ -259,24 +309,55 @@ def prox_uncentered_symmetry(X, step, center=None, strength=.5, use_soft=True, f
...
+          in scarlet that works for fractional pixel shifts.
+        * If `algorithm = "sdss" then the SDSS symmetry is used,
+          namely the source is made symmetric around the `center` pixel
+          by taking the minimum of each pixel and its symmetric partner.
+          This is the algorithm used when initializing an `ExtendedSource`
+          because it keeps the morphologies small, but during optimization
+          the penalty is much stronger than the gradient
+          and often leads to vanishing sources.
+        * If `algorithm = "soft" then soft symmetry is used,
```
herjy: How useful is `soft` now? 
Do we still have a need for treating perfectly centered sources differently?

scarlet/operator.py
```diff
@@ -259,24 +309,55 @@ def prox_uncentered_symmetry(X, step, center=None, strength=.5, use_soft=True, f
...
     result: `function`
         The update function based on the specified parameters.
     """
-    if use_soft:
+    if algorithm == "kspace" and (shift is None or np.all(shift == 0)):
+        algorithm = "soft"
+        strength = 1
+    if algorithm == "kspace":
+        return uncentered_operator(X, prox_kspace_symmetry, center, shift=shift, step=step, fill=fill)
+    if algorithm == "sdss":
```
herjy: `elif` ?

scarlet/source.py
```diff
@@ -358,7 +361,7 @@ def update(self):
 
 class ExtendedSource(PointSource):
     def __init__(self, frame, sky_coord, observation, bg_rms, thresh=1,
-                 symmetric=False, monotonic=True, center_step=5, delay_thresh=10, **component_kwargs):
+                 symmetric=True, monotonic=True, center_step=5, delay_thresh=10, **component_kwargs):
```
herjy: Why should point sources be made symmetric?

scarlet/update.py
```diff
@@ -214,13 +216,106 @@ def translation(component, direction=1, kernel=interpolation.lanczos, padding=3)
     return component
 
 
-def symmetric(component, strength=1, use_prox=True, kernel=interpolation.lanczos, padding=3):
+def psf_weighted_centroid(component):
+    # Extract the arrays from the component
+    morph = component.morph
+    if component.frame.psfs is None:
+        if not hasattr(component, "_centroid_psf"):
```
herjy: Hard coded shape??
I don't see why we should create a psf if the user did not specify one.

scarlet/update.py
```diff
@@ -214,13 +216,106 @@ def translation(component, direction=1, kernel=interpolation.lanczos, padding=3)
     return component
 
 
-def symmetric(component, strength=1, use_prox=True, kernel=interpolation.lanczos, padding=3):
+def psf_weighted_centroid(component):
+    # Extract the arrays from the component
```
herjy: The descriptor could be a little more explicit. Especially given that it returns a `component` and not arrays.

scarlet/operator.py
```diff
@@ -1,5 +1,6 @@
 from functools import partial
 
+from scipy import fftpack
```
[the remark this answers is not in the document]
fred3m: I needed it in my original implementation but removed the dependent.

scarlet/operator.py
```diff
@@ -248,7 +249,56 @@ def prox_soft_symmetry(X, step, strength=1):
...
+    """
+    # Record the morph shape
+    shape = X.shape
+    dy, dx = shift
+    padding = np.max(X.shape) + padding // 2
+    edges = ((padding, padding), (padding, padding))
+    corner = (padding, padding)
+    zeroMask = X <= 0
+    X[zeroMask] = 0
+    X = np.pad(X, edges, 'constant')
```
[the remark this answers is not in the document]
fred3m: Not in this case. It turns out that padding in `fftn` adds the zero rows/columns to the end of the array but this algorithm requires that the array to be made symmetric is centered, which is a bummer.

scarlet/operator.py
fred3m: It is faster to to the direct symmetry, since it's just `(x + x[::-1, ::-1])/2`, so we should use that if there isn't a shift. I don't necessarily see the need for the `strength` parameter but there doesn't seem to be any harm in keeping it.

scarlet/operator.py
fred3m: I *think* that it is more canonical to use another `if` when the previous `if` statement will always return from the function.

scarlet/source.py
fred3m: Testing shows that symmetry does improve deblending and this was the default scarlet setting before the API change, it was just turned off because it wasn't working properly. Now that it is I think we should return to the expected default.

scarlet/update.py
fred3m: This is really just Jim's function to weight the center of flux to give more weight to the central pixels. So in reality any function that falls off from the peak can be used, so think of `centroid_psf` as `centroid_weight`. Perhaps I should just rename it to make its use more clear.

scarlet/update.py
fred3m: Agreed. This was just copy pasted from Nate's code and I didn't pay too close attention to the (lack of a) docstring.

scarlet/operator.py
herjy: Right, I missed that

scarlet/source.py
herjy: I surrender to the impirical argument, but isn't it a little strange? Aren't point source supposed to be intrinsically symmetric?

scarlet/update.py
herjy: I am in favour of renaming. If I read this correctly, this `weight map` only serves to get the centroids?
If so, I am good with that.

herjy: Although, could it happen that `component.pixel_center` be way off? (e.g heavily blended scene)
In this case I am worried that this weigthing might bias the centering.

scarlet/source.py
fred3m: This is actually an `ExtendedSource`, which inherits from `PointSource`. And `PointSource` is actually a bit of a misnomer in that it is a source that is initialized as a point source but then is allowed to have its morphology updated. It is only a true point source if `fix_morph=True` (which might be a good idea to make the default behavior so that it really is a point source).

herjy: Yup! Completely misread that! Sorry.
I don't have anything against living a bit of freedom to point sources morphologies unless we know for sure something is a point source (star).

scarlet/update.py
fred3m: While I can't say for sure, I don't think so. This isn't a traditional centroid, in that the centroid is acting on the *model*, not the *image*. So while the image would see interference in the center of flux measurement from neighbors, after the first few iterations the model center of flux should be more stable. Also, `pixel_center` is the point about which we are making the solution monotonic so if it is way off then our entire model will be off.

scarlet/source.py
fred3m: It's actually even more complicated because even point sources should have their positions updated (since it is highly unlikely that they will be exactly centered). In fact, this should probably be an issue.

pmelchior:
Remy and I just had a debrief of Nate on the details of the algorithm. Two things came up.

* Nate uses a mask, which predetermines the source vs background footprints, and reapplies that mask after the symmetry operation to null the small artifacts that arise from the FFTs. Please make sure that this mask isn't called before the FFTs because it would generate unnecessary ringing. The FFT artifact may also make it harder for our histogram-based l0 sparsity to determine a threshold.

* The shift for the symmetry is only for fractional pixel offsets. The same approach should be able to do several pixel shifts, which should allow us to prevent to centering/passing operation that is currently implemented.

herjy:
zero masking is indeed applied before FFTs. As for the l0, there is also the option to set the threshold according to noise levels.

scarlet/operator.py
```diff
@@ -248,7 +248,56 @@ def prox_soft_symmetry(X, step, strength=1):
...
+    then transformed back to real space.
+    """
+    # Record the morph shape
+    shape = X.shape
+    dy, dx = shift
+    padding = np.max(X.shape) + padding // 2
+    edges = ((padding, padding), (padding, padding))
+    corner = (padding, padding)
+    zeroMask = X <= 0
+    X[zeroMask] = 0
```
herjy: Zero masking before the FFTs should go.

fred3m: Right, but we still want to save the mask to reassign the zeros at the end, right?

herjy: I think so. Otherwise, as @pmelchior said, the artifacts would mess with the threshold estimate when applying the l0. Although a preferable but still not ideal solution would probably be a smooth decrease rather than a sharp drop to zero as Nate pointed out.

herjy: Either way the mask is useful anyway.

fred3m:
> * The shift for the symmetry is only for fractional pixel offsets. The same approach should be able to do several pixel shifts, which should allow us to prevent to centering/passing operation that is currently implemented.

The current implementation of the centering algorithm does calculate a shift over multiple pixels and updates `pixel_center` (see https://github.com/fred3m/scarlet/blob/tickets/DM-20431/scarlet/update.py#L294). The problem is that to use the PSF for weighting, the algorithm needs a pixel center to start with. We *could* use the `pixel_center` calculated from the previous iteration but I think that using the pixel with the maximum flux in the current iteration is a better choice, since it is calculated after the gradient step.

scarlet/operator.py
fred3m: fixed

pmelchior:
That's not the point. I'm proposing that we don't go through the `uncentered_operator` method at all:
https://github.com/fred3m/scarlet/blob/tickets/DM-20431/scarlet/operator.py#L172
Instead, we can chose the shifting operator in Fourier space to do arbitrary translations directly in the original model frame.

fred3m:
I understand that after our telecon. I think that you're right, and we should probably create an issue to do just that.

scarlet/operator.py
```diff
@@ -248,7 +249,56 @@ def prox_soft_symmetry(X, step, strength=1):
...
+
+    # symmeterize
+    result_fft = result_fft.real
+
+    # Shift back
+    result_fft = result_fft*inv_shifter
+
+    # Transform to real space
+    result = np.fft.fftshift(np.fft.ifftn(result_fft))
+    # Return the unpadded transform
```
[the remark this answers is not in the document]
fred3m: I tested the code using `observation._centered` and it works the same, but that code has a few extra operations because the size of the padding is not known. Since it's only two lines of code I think it's best to leave as is, but if you feel strongly about duplicate code I can push a commit using `observation._centered`.

scarlet/source.py
```diff
@@ -333,23 +333,26 @@ def update(self):
             it = self._parent.it
         # Update the central pixel location (pixel_center)
         update.fit_pixel_center(self)
-        if it > self.delay_thresh:
-            update.threshold(self)
+        # Thresholding needs to be fixed (DM-10190)
```
pmelchior: So, we're running without thresholding and bboxes right now. How bad is that and what needs to get done to use it again?

scarlet/operator.py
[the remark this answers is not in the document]
herjy: I agree, it doesn't matter

scarlet/update.py
```diff
@@ -214,13 +216,115 @@ def translation(component, direction=1, kernel=interpolation.lanczos, padding=3)
...
+        morph = component.morph[bbox.slices]
+        shape = morph.shape
+        if shape[0] <= 1 or shape[1] <= 1:
+            return component
+        center = pixel_center[0]-bbox.bottom, pixel_center[1]-bbox.left
+    else:
+        morph = component.morph
+        shape = component.shape[-2:]
+        center = pixel_center
+    morph = morph.copy()
```
pmelchior: Why is there a copy here? Isn't the point of an update to overwrite the original?

scarlet/source.py
fred3m: We never really tested this algorithm for robustness and when running on the HSC patches during the sprint we saw a number of cases, primarily for resolved extended objects where the bulge and disk had very different magnitudes, where the thresholding algorithm would entirely wipe out the disk.

So we need to revisit this algorithm again (I have the opened the referenced JIRA ticket for this) and decide the best way to incorporate some information about the background in the calculation so that we don't cut too high. In the meantime I would recommend going with the attitude that "it's better to do things correctly than fast" and just skip thresholding until we can fix it and test it more rigorously.

scarlet/update.py
fred3m: I just copied the code from `monotonicity` that used the bounding box without thinking. I don't remember why we needed it there, I think it might have had something to do with trying to use proximal operators with `autograd` arrays when I first started out with them and there being some sort of conversion issue between `autograd` arrays and `Eigen` matrices that we use in monotonicity (`autograd.numpy.Array.copy` returns a numpy array, not an array with gradients... I think). I'll fix this and push the commit.

Issue #81, opened 2019-04-24
Convergence of faint objects


pmelchior:
This is for the new `fullscene` code:

The convergence test we perform is purely relative, i.e. the L2 norm of changes morphology/SED vectors need to be less than e.g. 1e-2 of the L2 norm of the vector itself, but that's hard to do if the vector is small to begin with.

The complication is that the normalization option tells us which vector is normalized to unity, making the other one small. But normalization is under user control, so we don't know that within `Blend`.

The second complication is that an absolute tolerance would have to be set in relation to the noise levels, but this is *very* hard to propagate from the observation to the model.

It would nonetheless be useful to check if there are meaningful changes from further iterations, so that we don't waste iterations in incremental improvements.

One option would be to check `Blend.mse` for changes and terminate if it doesn't improve by some amount, to be set by the user e.g. `Blend.fit(100, e_rel=1e-2, e_mse=1e-3)`.

PR #113, opened 2019-07-03
Ndim fft

herjy:
All right, I am mildly fed up with this, so I will submit it for review. The critical part, which is the ndimensional fft for `Observation` works marvelously well. The same trick does not work at all for `LowResObservation`. If anything it makes everything slower. Here is a recap of what you'll find:
* On the simple example from the notebook, the execution of `blend().fit` drops from 88 ms to 18 ms. I will do more tests to see hoe the various dimensions of the arrays affect this, but it looks nice! 
* Problem: It does not pass the tests and I am not entirely sure why. 
* `make_operator` is now a method of `LowResObsevation`, which makes more sense.
* I considered switching `resampling.match_patches` to a method of that class too, but its use in `match` and `match_psfs` made me think twice.
* I timed the setup time for `LowRes` and the run times for `blend.fit` in the notebooks. For comparison, the same thing is done on the current implementation in a branch I called 'speed_test' which I will use to investigate further the gain that the new implementation gives us. 
On the bad news side (yes this was all good news):
* The more I try to use Ndim_fft in `LowRes`, the slower it gets. I tried an implementation full-on ndim fft (it was beautiful, not a single for loop), but after running (building the resampling operator) for MANY minutes I lost patience and gave up. You will find a hybrid implementation that still uses scipy. 
* The prototypical code for ndim fft in `make_operator is left commented. I could never make it work for reasons that are still unclear. 

Since I already spent days on trying to make `LowRes` faster without success, Idon't want to delay this any further, I suggest that you guys review the part in `Observation`. If I can find why `LowRes` is slow, I'll update everything. Writing this makes me think that it could be a problem of `fftpack_shapes` but I am not convinced. 
Good luck.


herjy:
I also see in the tests that number of iterations are larger in the tests, which is weird given that they were exactly the same as the `master` branch in the notebook example.

pmelchior:
As for the number of iterations, this is possibly because the default behavior has the new symmetry switched on. I also saw slower convergence driven by a single faint object, so this is likely related to #81

herjy:
Yeah, but this would have shown in passing the tests in #81

scarlet/observation.py
```diff
@@ -144,7 +148,7 @@ def __init__(self, images, psfs=None, weights=None, wcs=None, channels=None,
             prevent artifacts from the FFT.
         """
         self.frame = Frame(images.shape, wcs=wcs, psfs=psfs, channels=channels)
-
+        self.psfs = psfs
```
pmelchior: this is a redundant reference. `psfs` is an attribute of `self.frame`

scarlet/interpolation.py
```diff
@@ -206,7 +206,7 @@ def outer(x):
     x = np.abs(dx - window)
     result = np.piecewise(x,
                           [x <= 1, (x > 1) & (x < 2)],
-                          [lambda x:inner(x), lambda x:outer(x)])
+                          [lambda x: inner(x), lambda x: outer(x)])
```
pmelchior: it looks like you switched on flake8. Lots of little changes, which we wanted to do in one go in a commit with no functionality change.

scarlet/observation.py
```diff
@@ -176,41 +180,50 @@ def match(self, model_frame):
...
+            kernels = np.fft.ifftshift(np.fft.irfftn(_psf_fft / target_fft, _fftpack_shape, axes=(1, 2)), axes=(1, 2))
+
+            if kernels.shape[1] % 2 == 0:
+                kernels = kernels[:, 1:, 1:]
+
+            kernels = _centered(kernels, psf_shape)
+
+            _diff_kernels_fft = np.fft.rfftn(kernels, self._fftpack_shape, axes=(1, 2))
 
             self._diff_kernels_fft = np.array(_diff_kernels_fft)
```
pmelchior: why is this copy necessary? you should be able to use the result of the `ffftn` in line 226.

scarlet/observation.py
```diff
@@ -219,8 +232,8 @@ def match(self, model_frame):
     def _convolve_band(self, model, diff_kernel_fft):
```
pmelchior: The method should be renamed to `_convolve`, and since `diff_kernel_fft` is an attribute of self, we can drop it from the argument list. This affects line 254.

scarlet/observation.py
```diff
@@ -274,22 +285,147 @@ def __init__(self, images, wcs=None, psfs=None, weights=None, channels=None, pad
         self.images = np.array(images)
         self._padding = padding
 
+        self.psfs = psfs
```
pmelchior: same things as above, `psfs` is in `self.frame`

scarlet/observation.py
herjy: Indeed

scarlet/interpolation.py
herjy: As I said, this is automated on my ide, but I thought it was pep8 I can check and change that if required.

pmelchior: Leave it, but switch off pep8 enforcement in your IDE for now. It makes the changes harder to read.

scarlet/observation.py
```diff
@@ -204,7 +204,7 @@ def match(self, model_frame):
             self.slices = tuple(([slice(s) for s in shape]))
 
             # Now we setup the parameters for the psf -> kernel FFTs
-            shape = np.array(model_frame.psfs.shape) - 1
+            shape = np.array(model_frame.psfs.shape)
```
pmelchior: can you explain this change?

scarlet/observation.py
```diff
@@ -451,7 +452,7 @@ def match(self, model_frame):
         self.slices = tuple([slice(s) for s in shape])
 
         # Now we setup the parameters for the psf -> kernel FFTs
-        shape = np.array(model_frame.psfs.shape) - 1
+        shape = np.array(model_frame.psfs.shape) #- 1
```
pmelchior: instead of commenting out the `-1`, please remove it

docs/quickstart.ipynb
```diff
@@ -188,11 +188,13 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "\n",
+    "import time\n",
```
pmelchior: is there a reason to replace `%timeit` here? Did you have any problems with the measurements?

docs/tutorials/multiresolution.ipynb
```diff
@@ -51,10 +53,10 @@
    "outputs": [],
    "source": [
     "# Load the HSC image data\n",
-    "obs_hdu = fits.open('../../data/test_resampling/Cut_HSC.fits')\n",
+    "obs_hdu = fits.open('../../data/test_resampling/Cut_HSC1.fits')\n",
     "data_hsc = obs_hdu[0].data.byteswap().newbyteorder()\n",
     "wcs_hsc = WCS(obs_hdu[0].header)\n",
-    "channels_hsc = ['g','r','i']\n",
+    "channels_hsc = ['r','i','z']\n",
```
fred3m: This makes more sense, it didn't look like a `gri` image to me.

docs/quickstart.ipynb
herjy: Yes, timeit runs the same function several time, which, in the case of scarlet does not make much sense since it updates some of the objects such that the second run starts as initialise by the result of the previous run. I just wanted to make sure that this was not causing anything weird with regard to the measurement. Also `%timeit` gives results like 80 ms, which is not what we experience. 12s, the result from time.clock is much more realistic.
Actually because %timeit repeats the same command we should not using on blend.fit because the first runtime will alway be MUCH larger than the following ones.

pmelchior: Then `%time` is your friend ...

scarlet/observation.py
pmelchior: @herjy ?

herjy: Sorry, as I said, I'm away. To be completely honest, I'm not entirely sure as to the fundamental reason here. All I know is that before I did that, the tests gave me a result that was shifted by 1 pixel exactly and now it doesn't. I strongly suspected that this was related to the introduction of the condition on pair-sized psfs. I wonder if this could be linked to our dipole problem.

pmelchior: Changes like that are very problematic because we don't understand why they're needed. Can you or @fred3m check if there was implicit padding done before or after this piece is called, which differs in the nD FFT and the 1D FFT.

Btw, I don't think our dipoles come from that: they are much smaller than 1-pixel shifts.

pmelchior:
Besides the three open comments, it would be nice to have a simple comparison of the time for `obs.match(frame)` (with PSF matching) and of `obs.render(model)` in three configurations of (C,Ny,Nx)
* 1, 1000,1000
* 5, 100, 100
* 1000, 100, 100

herjy:
Sorry, I saw your latest comment just now. I had similar stuffs done, so I'll do that quickly tomorrow. In the mean time, here is a new commit with Ndim_fft in place where it can be done without losing anything
I also added 5bands to the data, so that we have nicer colour images ;)

herjy:
On the performance side, here is what this PR changes. In this plot, the y-axis is the time gained in percents by using this implementation compared to the master branch at the time of the PR for three different tasks making use of `rfftn(axes = (1, 2))`: psf convolution `observation.render()`, observation setup `observation().match(frame)`, fitting `blend.fit()`. 
bands_gain

pmelchior:
OK, this looks good in the regime where we need it. The drop off at the right is probably because of memory thrashing / cache misses.

herjy:
Finally, I figured it out! So: for some values of the psf size, the fast shape of the fft returns an even value, which autograd.fft cannot deal with. To have autograd run smooth and clean, we tested the fast shape and made it uneven, but only along one axis, which, apparently, for the 2-d fft was ok, but not when going Ndim (this is still a little weird to me). So now I make both axes uneven whenever things fast shapes are even and we the Fourrier operations run nominally.

docs/tutorials/multiresolution.ipynb
```diff
@@ -245,26 +249,29 @@
     "residual_lr_rgb = scarlet.display.img_to_rgb(residual_lr[:,:-5])\n",
     "\n",
     "# Get the HR residual\n",
```
herjy: Just changed a few things to make plots look nicer.

fred3m:
Weird. It might be worth creating an issue with autograd or even fixing this ourselves upstream because I doubt that we're the only ones to get dinged by this.

scarlet/interpolation.py
```diff
@@ -317,7 +317,13 @@ def sinc_interp(coord_hr, coord_lr, sample_lr):
     hx = np.abs(x_lr[np.int(np.sqrt(np.size(x_lr))) + 1] - x_lr[0])
 
     assert hy != 0
```
herjy: The shape condition is useful in th case of the psf interpolation.

scarlet/observation.py
```diff
@@ -36,6 +39,7 @@ class Frame():
     channels: list of hashable elements
         Names/identifiers of spectral channels
     """
```
herjy: Ignore until line 186. format changes. I swear I did not do it on purpose :'D

pmelchior:
Is this a problem from autograd or from numpy FFTs?

fred3m:
It's a problem with autograd. Numpy.rfftn has no problem with even shapes, in fact I think that it prefers them and this code (without the check for an even last dimension) is the same basic algorithm that `scipy.signal.fftconvolve` uses.

But with autograd they raised an Exception when using rfftn with an even numbered last dimension, hence the fix that is in master (https://github.com/fred3m/scarlet/blob/master/scarlet/observation.py#L190-L194). But I had never run across a problem with PSF sizes, so I didn't notice that I needed to use the same fix when we calculated the optimal shape for PSF matching.

pmelchior:
So, it looks like we are in a good state now. @herjy please confirm that we can merge this PR.

herjy:
To answer Fred's last comment, It is somewhat strange that the problem did not arise earlier, but did now in the ndim. It could be related to the ndim implementation and the use of `axes`, but looking at the code for fft, I don't see  it. 
Anyway, I now have a much better grasp at this part of the code, and I think we are ready to merge.

pmelchior:
I just spotted this: cb8bd409 deletes the file `docs/changes.rst`. That's our change log and should not be delete. I'll restore it but future future reference: `git rm` needs to be used with caution!

Issue #114, opened 2019-07-04
only call `grad` once in `Blend`


pmelchior:
[Our use of `grad`](https://github.com/fred3m/scarlet/blob/5482878cf7f0847abc4a0b29ee55d58fd2b82eac/scarlet/blend.py#L121) is inefficient because it's called during every iteration even though the backward gradient tree doesn't change at all. Store the gradient function at the first call of `_backward` and call that function every time from then on.

fred3m:
You should just be able to set `self._grad = grad(self._loss, tuple(range(len(parameters))))` in `Blend.__init__`. Then just call `self._grad(*parameters)` in `self._backward`.

PR #115, opened 2019-07-04
Grad performance

pmelchior:
This PR addresses #114 

It does fix a lapse in our `Blend._backward` code, which iterated over sources instead of components. In essence, the code only worked for single-component sources. I've corrected that and removed the redundant `nodes` in favor of the more obvious `sources` property. I needed to adjust 2 sets of tests.

However, the main idea was to call `grad` only once instead of dynamically recomputing the gradient tree. This has *no* performance gains I could find with timeit for our user guide blend, running for 100 iterations. The take-away is that our runtime is fully dominated by computing the gradients.

### DIFF
--- a/scarlet/component.py
+++ b/scarlet/component.py
@@ -284,7 +284,7 @@ class ComponentTree():
     def sources(self):
         """Initial list of components or sources that generate the tree.
 
-        This will be different than `self.components` because sources can
+        This can be different than `self.components` because sources can
         have multiple components.
 
         Returns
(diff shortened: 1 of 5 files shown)


scarlet/component.py
```diff
@@ -281,8 +281,11 @@ def frame(self):
         return self._tree[0].frame
 
     @property
-    def nodes(self):
-        """Initial list that generates the tree.
+    def sources(self):
```
herjy: Since you changed this, maybe the `n_nodes` property should be changed to `n_sources`

Issue #121, opened 2019-07-09
tests should check for functionality not structure


pmelchior:
The current unit testing idea is centered too strongly around testing that every attribute of classes is set right. This leads to long adjustment for changes to the code base.

Instead, we need to identify for every class all functions and functionalities we rely on, and test that they are doing what they are supposed to. For instance, how components organize their internal structures is largely irrelevant, what matters is that given some SED and morphology, they produce an `autograd`-compatible model.

A good example of functionality testing is for [`update`](tests/test_update.py) methods, unnecessarily introspective ons are virtually all `test_init` cases, which one test if the arguments of the init call are stored properly. Because they require exact names, a simple internal renaming of a variable causes these tests to fail.

This issue has two parts:
* remove all structural tests
* add functionality tests for every outward facing method

What is not part of this issue is performance and robustness testing on actual data.

pmelchior:
Closed by #147

Issue #111, opened 2019-07-02
Choose fast sizes for FFT in Blend frame


pmelchior:
The speed of FFTs is strongly dependent on the size of the arrays. The reference frame should be chosen to have a number of pixels as a power of 2.

The minimal solution would be to log a warning if a `Frame` instance does not have a proper size. However, data frames don't need that because they'll never be FFTed. One could add an attribute to `Frame` to specify that it's for a model instead of data, but that's clunky. One could instead issue the warning if a `Component` is created with an improper frame, but then it's too late and the warning would be raised every time for k components. Raising it in `Blend` is still too late.

I still think the first option is best: raising a warning whenever a Frame is created. It would also prevent someone from using a data frame for the model.

fred3m:
It's not quite that simple. See https://docs.scipy.org/doc/scipy/reference/generated/scipy.fftpack.next_fast_len.html . In short you are looking for the nearest factor of 2, 3, or 5, since those are the primes that fftpack (which is what numpy uses) are optimized for. This can be 100's of times faster, as they describe in the link.

pmelchior:
This is quite a rabbit hole. It's easy to issue a warning that the size isn't well-chosen, but if the user acts on it and changes the size of the blend frame, we have a size mismatch.
The init functions for several sources and the loss of `Observation` all assume that the data and the mode frame have the same size. This relation would be broken.

Given problems with padding and objects at the edges, we will probably have to consider slices to match the spatial footprints of model and data frame. In `LowResObservation` we already have one of those slices (in data space), but all observations need both slices (in model space if the model encompasses the data; and in data space if the data encompasses the model). This would allow us to grow the model explicitly to reduce problems at the edges. 

This issue does not deal with how to pick the frame sizes given a set of sources and observations, but it should enable us to be flexible about it.

pmelchior:
In #124 we thought that we don't need the resizing of the Blend frame because the model is padded to a fast size anyway before it is convolved. But the symmetry update still operates on the model frame as is, which makes this operation still inefficient. We might be able to close this issue with improvements from #110

pmelchior:
Since we're not using k-space symmetry and we use fast sizes in `FFT`, this issue is largely mute.

Issue #109, opened 2019-07-01
Use standard numpy for all modules that don't need autograd


pmelchior:
Only `Blend` and `Observation` need `autograd.numpy`. We should avoid using it in all modules that are not in the backward path because it generates overhead and some vanilla numpy methods may be missing in autograd.

Issue #106, opened 2019-06-28
Use consistent dtypes


fred3m:
We are not guaranteed that users will always use the default double precision for the input images and the output model (eg. LSST uses single precision). Internally we occasionally use `numpy.zeros`, `numpy.ones`, and perhaps other functions that generate an array of type `np.float64`, which results in unnecessary casting operations and can even give output models of the wrong type.

We should go back through the code, make sure that we are always using the proper `dtype`, and add an assert to nearly all of our tests to make sure that we are getting the expected output types.

pmelchior:
Agreed. I also came to this problem with my `RandomSource` code which doesn't even require an observation and thus has no way of know what the `dtype` is.

How about adding `dtype` to `Frame`?

fred3m:
I think that's a good idea. We can have it default to `numpy.float64` since that is the numpy default.

pmelchior:
`np.float32` should do and give us 2x memory

PR #124, opened 2019-07-11
Cleanup

pmelchior:
This PR addresses issues #106, #109, #111, and a bit of #121. No feature changes, just consistency and  avoidance of type casting.


scarlet/observation.py
```diff
@@ -35,10 +35,16 @@ class Frame():
...
+        Dtype to represent the data.
     """
-    def __init__(self, shape, wcs=None, psfs=None, channels=None):
+    def __init__(self, shape, wcs=None, psfs=None, channels=None, dtype=np.float32):
         assert len(shape) == 3
         self._shape = tuple(shape)
+
+        if self.Ny != fftpack.next_fast_len(self.Ny) or self.Nx != fftpack.next_fast_len(self.Nx):
+            logger.warning('Shape ({},{}) not fast for FFT. Consider increasing to ({},{}) pixels'.format(self.Ny, self.Nx, fftpack.next_fast_len(self.Ny), fftpack.next_fast_len(self.Nx)))
+
```
fred3m: I might be missing something but I don't think that this is necessary. The shape that we want to be fast is the shape that is `Frame.shape + Observation.psfs.shape`, since we need padding to prevent artifacts in the image. So the user shouldn't really need to care about this at all, since we'll be padding anyway.

tests/test_component.py
```diff
@@ -78,10 +36,8 @@ def test_methods(self):
         assert_array_equal(model, truth)
         assert_array_equal(other_model, other_truth)
 
-        with pytest.raises(ValueError):
-            component.get_model(sed)
-        with pytest.raises(ValueError):
-            component.get_model(morph=morph)
+
+        # TODO: test component with partial parameter list
```
fred3m: Typically on LSST anytime we place a `TODO` in the code we open a ticket and put the ticket number in the comment. I think that's a good practice for us to adopt, so it might be worth opening an issue and writing the issue number below the `TODO` comment.

scarlet/observation.py
pmelchior: Hmm, right. Does that mean #111 is a non-issue?

fred3m: Yeah, either we were being more clever last week than I am right now or I wasn't thinking about this properly when that issue was created.

PR #131, opened 2019-08-30
Fast resampling

herjy:
A fast implementation of the resampling and convolution that works for rotated frames. 
The trick used on the non-rotated frames in PR #130 was a particular case of a more generic formula that can in general be used to perform convolutions and/or interpolations (it's the same anyway) with/on  rotated frames/images. It essentially comes from the separation of the  axes of the rotated frames through the application of the rotation matrix. In theory, this exact same trick should be useful for any stationary affine transformation of the grid. 

This trick makes the operation ~ twice as slow as in the non-rotated case, but that is because we have to perform ~ twice as many operations. This is still an order of magnitude faster than doing it the stupid way.

I also solved an issue in the image matching by realising that pixels have a width and by setting a pixel's coordinate to the bottom left corner of the pixel.


scarlet/observation.py
```diff
@@ -558,21 +670,44 @@ def _render(self, model):
...
-        for c in range(self.frame.C):
+        if self.isrot:
+            if self.small_axis:
+                model_conv1d = sinc_shift_2D(model_, [-self.sin_rot * self.X_unrot, self.cos_rot * self.X_unrot],
+                                             axes=[None], doit = 1)
+                model_shape = np.shape(model_conv1d)
+                for c in range(self.frame.C):
+                    model_image.append(np.dot(model_conv1d[c].reshape(model_shape[1],
+                                                                      model_shape[3] * model_shape[2]),
+                                              self._resconv_op[c].T).T[:, ::-1])
```
herjy: I can totally explain the axes inversions and transposes.

herjy:
A strange behaviour in this PR:
I now cache part of the operator that performs shifting for `prox_kspace_symmetry`. namely the values 
```shift_y = np.exp(-1j * 2 * np.pi * freq_y)
        shift_x = np.exp(-1j * 2 * np.pi * freq_x)```
I compared this strategy with the master branch with an extra twist: the shifting can be done by computing the outer product of the shifting operators in both direction 
```X_fft * np.outer(shifter_y ** (-dy), shifter_x ** (-dx))```
or by broadcasting the shifts in each direction alternatively. This is a comparative timing of both of these strategies compared with `master` branch. 
While the broadcasting operation does not seem faster or slower (`render` operations take the same amount of time), the overall run faster, which is to be expected from caching. However, this is not observed when using the outer product, which makes no sense to me. 
<img width="440" alt="Screen Shot 2019-09-24 at 4 24 18 PM" src="https://user-images.githubusercontent.com/16084926/65547696-0047df80-dee8-11e9-892a-279cf1465bf0.png">
<img width="446" alt="Screen Shot 2019-09-24 at 4 25 11 PM" src="https://user-images.githubusercontent.com/16084926/65547697-0047df80-dee8-11e9-966a-49ea71dbf392.png">

pmelchior:
What's measured by `run`? Is that a single scarlet iteration?

herjy:
Like in previous PR, `run` is the time taken by a full execution of `fit`.

pmelchior:
I'm worried about the substantial runtime *increase* for a small number of bands, which is the regime we normally work in. What was the number of pixels in the tests in the bottom plots?

herjy:
It worries me too, BUT render being unaffected, I suspect there must be a reason that we can figure out and work out. Right after I finish my PR I will do some deeper profiling to address this.

pmelchior:
Can it be timing variance? How many different runs do you have for each of the points on the plot.

herjy:
It is only 7 runs of 1 loop each for `trun`, but I repeated the experiment and it is invariably slower.

PR #132, opened 2019-09-03
Tickets/DM-20127: Improve initialization

fred3m:
This ticket does two things:

* Implements a new `Fourier` class that stores an image and FFTs for any shape it has already calculated, making PSFs more useful and streamlining the psf matching code. It also fixes centering bug that can occur when performing operations in Fourier space
* Estimates the kernel to deconvolve the observations in order to make the initial morphology guess more accurate.

### DIFF
--- a/scarlet/fft.py
+++ b/scarlet/fft.py
@@ -1,6 +1,6 @@
 import operator
 
-import numpy as np
+import autograd.numpy as np
 from scipy import fftpack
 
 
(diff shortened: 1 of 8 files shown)


fred3m:
In the 10 scenes that I regularly test the new initialization saves around 10 iterations (for harder blends, there is less of a change for easier ones since they already converge in 10-15 iterations).

fred3m:
#161 produces better results using a deconvolution in real space that is free of many of the numerical artifacts contained in the FFT implementation.

PR #134, opened 2019-09-17
Add Fourier class

fred3m:
Add `Fourier` class to perform FFT and inverse FFT operations.


scarlet/interpolation.py
```diff
@@ -419,3 +419,52 @@ def sinc2D(y, x):
...
+    _y = np.linspace(y[0]-dy/2, y[-1]+dy/2, len(y)*dNy+1)
+    return f(_y, _x), _y, _x
+
+
+def apply_2D_trapezoid_rule(y, x, f, dNy, dNx=None, dy=None, dx=None):
+    """Use the trapezoid rule to integrate over a subsampled function
+    2D implementation of the trapezoid rule.
+    See `apply_trapezoid_rule` for a description, with the difference
+    that `f` is a function `f(y,x)`, where we note the c ++`(y,x)` ordering.
+    """
```
herjy: A miny tiny little comment. I would prefer if we put operations over `y` berfore the corresponding operation over `x`, just to remain consistent with the convention. I found the ordering helped avoiding confusion.

scarlet/interpolation.py
```diff
@@ -419,3 +419,52 @@ def sinc2D(y, x):
...
+    # Calculate the volume of each sub region
+    dz = 0.4 * (z[:-1, :-1] + z[1:, :-1] + z[:-1, 1:] + z[1:, 1:])
+    volumes = dx * dy * dz / dNy / dNx
+
+    # Sum up the sub regions around each point to
+    # give it the same shape as the original `(y,x)`
+    _dNx = len(_x) // dNx
+    _dNy = len(_y) // dNy
+    volumes = np.array(np.split(np.array(np.split(volumes, _dNx, axis=1)), _dNy, axis=1)).sum(axis=(2, 3))
+    return volumes
```
herjy: I realised that in general, trapezoid rule will be more accurate than rect convolutions unless we can do the convolution analytically or we take enough samples of the function. However, I wonder if the case might arise that we get non-analytic PSF that are not integrated, in which case, the rect convolution would apply?

scarlet/psf.py
```diff
@@ -10,33 +13,79 @@ def moffat(coords, y0, x0, amplitude, alpha, beta=1.5):
...
+    Returns
+    -------
+    result: array
+        A 2D circular gaussian sampled at the coordinates `(y_i, x_j)`
+        for all i in `y` and j in `x`.
     """
-    Y, X = coords
+    if amplitude is None:
+        amplitude = 1/(np.pi**2*sigma**2)
+    X, Y = np.meshgrid(x, y)
```
herjy: It doesn't really matter here, but for consistency `Y, X = np.meshgrid(x,y)`

scarlet/psf.py
```diff
@@ -74,9 +123,9 @@ def fit_target_psf(psfs, func, init_values=None, extract_values=None):
     """
     from scipy.optimize import curve_fit
 
-    X = np.arange(psfs.shape[2])
-    Y = np.arange(psfs.shape[1])
-    X, Y = np.meshgrid(X, Y)
+    x = np.arange(psfs.shape[2])
+    y = np.arange(psfs.shape[1])
+    X, Y = np.meshgrid(x, y)
```
herjy: `Y, X` or am I missing something?

scarlet/psf.py
```diff
@@ -95,7 +144,8 @@ def fit_target_psf(psfs, func, init_values=None, extract_values=None):
     def reshape_func(*args):
         """Flatten the function output to work with `scipy.curve_fit`
         """
-        return func(*args).reshape(-1)
+        _args = [y, x] + list(args[1:])
```
herjy: I'm not sure I follow. Can you explain this?

scarlet/source.py
```diff
@@ -390,11 +390,11 @@ def __init__(self, frame, sky_coord, observation, symmetric=True, monotonic=True
         if self.symmetric:
             if self.frame.psfs is None:
                 shape = (41, 41)
-                psf = generate_psf_image(gaussian, shape, amplitude=1, sigma=.9)
+                psf = generate_psf_image(gaussian, shape, amplitude=1, sigma=.9, normalize=False)
```
herjy: Why not normalising?

tests/test_blend.py
```diff
@@ -29,13 +29,14 @@ def init_data(shape, coords, amplitudes=None, convolve=True, dtype=np.float32):
...
+        target_psf = scarlet.psf.generate_psf_image(scarlet.psf.gaussian, psf_shape, sigma=.9)
         target_psf /= target_psf.sum()
 
-        psfs = np.array([scarlet.psf.generate_psf_image(scarlet.psf.gaussian, psf_shape, psf_center,
-                                                        amplitude=1, sigma=1+.2*b) for b in range(B)], dtype=dtype)
+        psfs = np.array([
+            scarlet.psf.generate_psf_image(scarlet.psf.gaussian, psf_shape, sigma=1+.2*b)
+            for b in range(B)
+        ], dtype=dtype)
+        psfs /= psfs.max(axis=(1, 2))[:, None, None]
```
herjy: What is the max normalisation for here?

scarlet/psf.py
```diff
@@ -10,33 +13,79 @@ def moffat(coords, y0, x0, amplitude, alpha, beta=1.5):
...
-    return func(coords, center[0], center[1], *args, **kwargs)
+    ry, rx = np.array(shape) // 2
+
+    y = np.linspace(-ry, ry, shape[0])
+    x = np.linspace(-rx, rx, shape[1])
+    f = partial(func, **kwargs)
+    result = apply_2D_trapezoid_rule(y, x, partial(f, **kwargs), subsamples)
+    if normalize:
+        result /= result.sum()
+    return result
```
herjy: Should we not make all psfs Fourier objects?

tests/test_blend.py
```diff
@@ -88,16 +90,17 @@ def test_fit_point_source(self):
         blend.fit(10)
 
         assert blend.it == 2
-        assert_almost_equal(blend.L_sed, 2.5481250470053265)
-        assert_almost_equal(blend.L_morph, 9024.538938935855)
+        #assert_almost_equal(blend.L_sed, 2.5481250470053265)
+        #assert_almost_equal(blend.L_morph, 9024.538938935855)
```
herjy: Remove?

tests/test_blend.py
```diff
@@ -115,7 +118,7 @@ def test_fit_extended_source(self):
 
         # Fit the model
         blend.fit(100)
-        assert blend.it == 13
+        assert blend.it == 15
```
herjy: I find these hard-coded numbers a little disturbing. I am pretty sure these are going to change everytime we change the algorithm. Maybe we could put `assert blend.it < 16` or 20 just to be on the safe side. This way, if a new minimisation makes convergence faster by even just one iteration, it won't crash the tests. `16` is a good way to make sure that further alterations of the code will only improve convergence.

scarlet/fft.py
```diff
@@ -0,0 +1,297 @@
...
+            axes = [axes]
+        for a, axis in enumerate(axes):
+            dS = newshape[a] - arr.shape[axis]
+            startind = (dS+1) // 2
+            endind = dS - startind
+            pad_width[axis] = (startind, endind)
+    return np.pad(arr, pad_width, mode="constant")
+
+
+def _get_fft_shape(img1, img2, padding=3, axes=None):
```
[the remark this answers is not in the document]
fred3m: Way back when we first started to doing FFTs I did some testing using different padding and it seemed like adding the two shapes with an additional 3 pixels provided the best pixel area to suppression of artifacts ratio, so I've been using it as the default padding ever since. I don't remember the exact number, but I do recall it saved a few orders of magnitude in the residuals over no padding.

scarlet/fft.py
```diff
@@ -0,0 +1,297 @@
...
+    else:
+        shape = np.zeros(len(axes))
+        try:
+            len(axes)
+        except TypeError:
+            axes = [axes]
+        for n, ax in enumerate(axes):
+            shape[n] = shape1[ax] + shape2[ax]
+
+    shape += padding
```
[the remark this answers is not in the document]
fred3m: see previous comment.

scarlet/fft.py
```diff
@@ -0,0 +1,297 @@
...
+        # Create views into the fft transformed values, appropriately adjusting
+        # the shapes for the new axes
+        fft_kernels = {
+            tuple([s for idx, s in enumerate(shape) if self.axes[idx] not in removed]): kernel[index]
+            for shape, kernel in self._fft.items()
+        }
+        return Fourier(self.image[index], fft_kernels, axes=axes)
+
+
+def _kspace_operation(image1, image2, padding, op, shape):
```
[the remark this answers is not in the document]
fred3m: Yes that's true and if you prefer I can change the name. I still think of the `Fourier` instance as an image, since the class is really a wrapper of an image that includes convenience functions to operate on its Fourier transform. But I can see how this might be confusing.

scarlet/interpolation.py
fred3m: Perhaps. If so we can just implement a rect convolution method in `interpolation` and either pass an option to `psf.generate_psf_image` to choose which type of integration to use or create a `psf.generate_psf_image_from_rect`. But I would save doing that until we discover that we need it.

scarlet/fft.py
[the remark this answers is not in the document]
herjy: I don't mind the name all that much, but it should be clearer in the docstring, I think.

scarlet/interpolation.py
herjy: I agree. Plus I'm not even sure the case might arise

scarlet/psf.py
fred3m: This is actually a flaw in numpy. I believe that the only numpy function that requires `(x, y)` ordering is the `meshgrid` function, since the order matters here (`X` has `len(y)` rows, where each row is `x` and `Y` has `len(x)` columns, where each columns is `y`).

scarlet/psf.py
fred3m: see previous comment.

scarlet/psf.py
fred3m: This is to match the API of `scipy.curve_fit`, which requires the dependent variable as the first argument and the result to be a 1D array. The old version of the psf functions passed the meshgrid `(Y, X)` but for the trapezoid rule it is more convenient to use a single 1D array for `y` and another for `x` instead of a mesh grid, so I had to create a hack to make the `curve_fit` function work properly.

scarlet/source.py
fred3m: This part of the code will hopefully be obsolete soon, since it doesn't really work as well as we had hoped, but the PSF is only used to weight the centroid. So the PSF should have a value of 1 for the center pixel, which is achieved in the next line so there didn't seem to be a need to perform an extra normalization in `generate_psf_image` that will be undone anyway.

tests/test_blend.py
fred3m: I like this idea. By the end of the year this should be removed completely because I hope to have QA testing in LSST by then so we'll be able to track exactly how long blends are taking on a much larger (and more useful) scale. In the meantime I like your proposed solution.

fred3m: I decided to go with 20, since there is a possibility that a future improvement (like deconvolution) might make this one example slightly worse while improving convergence in general.

tests/test_blend.py
fred3m: Yeah. I left it commented out to leave it open for discussion during review but I agree with Peter's sentiment from a few weeks ago that we should eliminate tests for internal things that the users don't see, and these parameters are so dependent on the internal workings of the code that the tests are essentially useless.

scarlet/psf.py
fred3m: Hmmm, you make a good point. I can't seem to think of a time that we'll want to generate a PSF that isn't a `Fourier` object and even if such a case comes up we can just call `Fourier.image` to get the image array anyway. This will take a little time updating a lot of the docs and tests but I think it is the correct thing to do.

tests/test_blend.py
fred3m: It's a way to scale the image such that the SED of each source, as detected on the center pixel of each source in `images` is the same as the SED used to initialize the image. So I make use of this particular normalization in all of the tests to ensure that I get out the same SED that I put in.

scarlet/fft.py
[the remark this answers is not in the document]
fred3m: Sure, no problem. I usually try to keep the docstrings light for private functions that are not part of the public API but in this case I think it is important to make it clear what is expected.

scarlet/psf.py
herjy: There is not a BIG advantage to have psfs turned into `Fourier` object so if the change is a hastle, maybe  it's not worth it.
Though, it would make declaring analytic psfs slightly nicer on the users side, instead of having to write something like `generate_psf_image(analytic_psf)`, we would just declare the analytic psf as a Fourier object. I think it also makes sense conceptually, but that sounds like the kind of stylistic detail @pmelchior likes to weigh in.

scarlet/psf.py
herjy: I had to check to convince myself of this, but I agree now. That is weird and 🙄

scarlet/psf.py
fred3m: Ok, there is an implementation decision that needs to be made if we return a `Fourier` object. The issue is that `generate_psf_image` returns a 2D PSF image, but the model image is a PSF with an added dimension for the band. This is accomplished by adding an axis to the PSF image using either `psf[None]` or `psf[np.newaxis]` (they are the same thing, since apparently `np.newaxis` is `None`).

The problem is if we add a new axis to a `Fourier` object via slicing we need to decide the default behavior, is it an axis that will be transformed or will it be an axis the the FFT skips? Currently it is up to the user to decide since they have to first add the new axis to the PSF image and then create a `Fourier` object where they specify the `axes` used.

My first thought is that if we add a new axis it should *not* be transformed, assuming that most people wanting to add an axis are adding a band dimension. But I could be missing something, so I'm curious to hear your thoughts.

fred3m: I went with my solution to skip any new axes and fixed the tests accordingly. Have a look @herjy and if you and @pmelchior  approve this ticket is ready to be merged.

PR #153, opened 2019-12-22
pep8 compliant code from black

### LINKED ISSUE #145: Use black for automatic PEP8-compliant code
[Black](https://black.readthedocs.io/en/latest/index.html) work really well with minimal intervention from users. I suggest that we use the git integration via [pre-commit](https://pre-commit.com/).

This *must* wait until we merged `adam` so to not confuse meaningful changes with stylistic changes.

pmelchior:
This PR has no functionality change. It reformats the code to PEP8 by using [black](https://black.readthedocs.io/en/latest/index.html). I also added an automated pre-commit hook to maintain PEP8-compliance automatically. Install instructions are [here](https://black.readthedocs.io/en/latest/version_control_integration.html)

### DIFF
--- a/scarlet/.pre-commit-config.yaml
+++ b/scarlet/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: stable
+    hooks:
+      - id: black
+        language_version: python3.7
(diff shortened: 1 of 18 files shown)


pmelchior:
Black's default line length is 88 chars (yeah, I know, everyone thinks theirs is best). It can be configured, but it's recommended not to. Anyway, black looks for config in a `pyproject.toml` file, and some editors might even respect that. I don't think we need to bother. As long as our editors don't "correct" black'ed code, you can code anyway you want and have the commit hook do the formatting. That's what I like about it.

PR #154, opened 2019-12-23
made BBox any-dimensional

pmelchior:
BBox was inelegant because it implicitly assumed or manually forced all boxes to be in 3D. Now it works in any dimension and simply takes its dimensionality from the contained that defined on at init. The only downside is the loss of the `front`, `back`, `left`, `right` etc. properties. On the upside, as Fred and I can attest: there isn't universal agreement on what they mean, so I've chosen the python slicing syntax `start` and `stop` and made it multi-dimensional.

### DIFF
--- a/scarlet/bbox.py
+++ b/scarlet/bbox.py
@@ -11,22 +11,17 @@ class Box:
     Parameters
     ----------
     shape: tuple
-        Size of the box in depth,height,width
+        Size of the box
     origin: tuple
-        Minimum (z,y,x) value of the box (front low left corner).
+        Front/low/left corner coordinate of the box
     """
 
-    def __init__(self, shape, origin=(0, 0, 0)):
-        # bbox always in 3D
-        if len(shape) == 2:
-            shape = (0, *shape)
-        assert len(shape) == 3
-        self.shape = shape
-
-        if len(origin) == 2:
-            origin = (0, *origin)
-        assert len(origin) == 3
-        self.origin = origin
+    def __init__(self, shape, origin=None):
+        self.shape = tuple(shape)
+        if origin is None:
+            origin = (0,) * len(shape)
+        assert len(origin) == len(shape)
+        self.origin = tuple(origin)
 
     @staticmethod
     def from_image(image):
@@ -35,7 +30,6 @@ class Box:
         Parameters
         ----------
         image: array-like
-            2D image
 
         Returns
         -------
@@ -45,34 +39,24 @@ class Box:
         return Box(image.shape)
 
     @staticmethod
-    def from_bounds(front, back, bottom, top, left, right):
+    def from_bounds(*args):
         """Initialize a box from its bounds
 
         Parameters
         ----------
-        bottom: int
-            Minimum in the y direction.
-        top: int
-            Maximum in the y direction.
-        left: int
-            Minimum in the x direction.
-        right: int
-            Maximum in the x direction.
+        args: ints
+            Min/Max coordinate for every dimension
 
         Returns
         -------
         bbox: :class:`scarlet.bbox.Box`
             A new box bounded by the input bounds.
         """
-        if back < front:
-            back, front = front, back
-        if top < bottom:
-            top, bottom = bottom, top
-        if right < left:
-            right, left = left, right
-        return Box(
-            (back - front, top - bottom, right - left), origin=(front, bottom, left)
-        )
+        assert len(args) % 2 == 0
+        dims = len(args) // 2
+        shape = [args[dim * 2 + 1] - args[dim * 2] for dim in range(dims)]
+        origin = [args[dim * 2] for dim in range(dims)]
+        return Box(shape, origin=origin)
 
     @staticmethod
     def from_data(X, min_value=0):
@@ -88,7 +72,7 @@ class Box:
         Returns
         -------
         bbox: :class:`scarlet.bbox.Box`
-            Bounding box for the thresholded `X` (bottom, top, left, right)
+            Bounding box for the thresholded `X`
         """
         sel = X > min_value
         if sel.any():
@@ -97,20 +81,16 @@ class Box:
             for dim in range(len(X.shape)):
                 bounds.append(nonzero[dim].min())
                 bounds.append(nonzero[dim].max() + 1)
-            if len(X.shape) == 2:
-                bounds.insert(0, 0)
-                bounds.insert(1, 0)
         else:
-            bounds = [0] * 6
+            bounds = [0,] * len(X.shape) * 2
         return Box.from_bounds(*bounds)
 
     def contains(self, p):
         """Whether the box cotains a given coordinate `p`
         """
-        if len(p) == 2:
-            p = (0, *p)
+        assert len(p) == self.D
 
-        for d in range(len(self.shape)):
+        for d in range(self.D):
             if p[d] < self.origin[d] or p[d] > self.origin[d] + self.shape[d]:
                 return False
         return True
@@ -125,27 +105,17 @@ class Box:
 
         Returns
         -------
-        If shape is 2D: `slice_y`, `slice_x`
-        If shape is 3: `slice(None)`, `slice_y`, `slice_x`
+        slices for every dimension
         """
         if hasattr(im_or_shape, "shape"):
             shape = im_or_shape.shape
         else:
             shape = im_or_shape
-        assert len(shape) in [2, 3]
+        assert len(shape) == self.D
 
         im_box = Box(shape)
         overlap = self & im_box
-        zslice, yslice, xslice = (
-            slice(overlap.front, overlap.back),
-            slice(overlap.bottom, overlap.top),
-            slice(overlap.left, overlap.right),
-        )
-
-        if len(shape) == 2:
-            return yslice, xslice
-        else:
-            return zslice, yslice, xslice
+        return tuple(slice(overlap.start[d], overlap.stop[d]) for d in range(self.D))
 
     def extract_from(self, image, sub=None):
         """Extract sub-image described by this bbox from image
@@ -164,10 +134,7 @@ class Box:
         imbox = Box.from_image(image)
 
         if sub is None:
-            if len(image.shape) == 3:
-                sub = np.zeros(self.shape)
-            else:
-                sub = np.zeros(self.shape[1:])
+            sub = np.zeros(self.shape)
         subbox = Box.from_image(sub)
 
         # imbox now in the frame of this bbox (i.e. of box)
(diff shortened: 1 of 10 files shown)


scarlet/bbox.py
```diff
@@ -11,22 +11,17 @@ class Box:
     Parameters
     ----------
     shape: tuple
-        Size of the box in depth,height,width
+        Size of the box
     origin: tuple
-        Minimum (z,y,x) value of the box (front low left corner).
+        Front/low/left corner coordinate of the box
```
herjy: This is still 3-dimensional. In the previous docstring there at least was a notion of minimum, which ignores the number of dimensions.

scarlet/bbox.py
```diff
@@ -11,22 +11,17 @@ class Box:
...
-        if len(origin) == 2:
-            origin = (0, *origin)
-        assert len(origin) == 3
-        self.origin = origin
+    def __init__(self, shape, origin=None):
+        self.shape = tuple(shape)
+        if origin is None:
+            origin = (0,) * len(shape)
+        assert len(origin) == len(shape)
+        self.origin = tuple(origin)
```
herjy: That's nice and cleaner, although I think I raised it at the previous PR, why would we need anything else but 3-D? Already 2D is, I think unecessary if we specify that everything has to be 3-D

scarlet/bbox.py
```diff
@@ -45,34 +39,24 @@ def from_image(image):
...
-        bottom: int
-            Minimum in the y direction.
-        top: int
-            Maximum in the y direction.
-        left: int
-            Minimum in the x direction.
-        right: int
-            Maximum in the x direction.
+        args: ints
+            Min/Max coordinate for every dimension
```
herjy: Imagine we have a lot of dimensions, the number of arguments becomes confusing. How about asking for `N` tuples and check that `N == len(self.shape)`

scarlet/bbox.py
```diff
@@ -45,34 +39,24 @@ def from_image(image):
...
-            back, front = front, back
-        if top < bottom:
-            top, bottom = bottom, top
-        if right < left:
-            right, left = left, right
-        return Box(
-            (back - front, top - bottom, right - left), origin=(front, bottom, left)
-        )
+        assert len(args) % 2 == 0
+        dims = len(args) // 2
```
herjy: SHouldn't we check `dims == len(self.shape)`?

scarlet/source.py
```diff
@@ -144,10 +144,10 @@ def trim_morphology(sky_coord, frame, morph, bg_cutoff, thresh):
...
             size = 2 * max(
                 (
-                    pixel_center[0] - bbox.bottom,
-                    bbox.top - pixel_center[0],
-                    pixel_center[1] - bbox.left,
-                    bbox.right - pixel_center[1],
+                    pixel_center[0] - bbox.start[0],
+                    bbox.stop[0] - pixel_center[0],
+                    pixel_center[1] - bbox.start[1],
+                    bbox.stop[1] - pixel_center[1],
```
herjy: This is the kind of place that makes me wonder why we want to generalise the dimensionality of the boxes. Here we consider that pixels are in dimensions `(0,1)` (btw, I thought it would be along `(1,2)` or `(-2,-1)`) and it makes me think that we have to think of what those other axes might be and we take them into account.

scarlet/bbox.py
fred3m: This is a static constructor, there is no `self.<anything>` yet.

scarlet/bbox.py
```diff
@@ -202,58 +169,22 @@ def insert_into(self, image, sub):
...
-        """Maximum y value
-        """
-        return self.origin[1] + self.shape[1]
-
-    @property
-    def right(self):
-        """Maximum x value
-        """
-        return self.origin[2] + self.shape[2]
+        return tuple(o + s for o, s in zip(self.origin, self.shape))
```
fred3m: It's probably worth making `Box` immutable and caching constants like this that will be used frequently.

scarlet/bbox.py
```diff
@@ -293,18 +223,12 @@ def __and__(self, other):
...
-    def __str__(self):
-        return "Box({0}..{1}, {2}..{3}, {4}..{5})".format(
-            self.front, self.back, self.bottom, self.top, self.left, self.right
-        )
+        assert other.D == self.D
+        bounds = []
+        for d in range(self.D):
+            bounds.append(max(self.start[d], other.start[d]))
+            bounds.append(min(self.stop[d], other.stop[d]))
+        return Box.from_bounds(*bounds)
```
fred3m: We should check that the minimum bound is less than the maximum bound. It might be that the sources don't overlap at all, in which case we should either raise an `Exception`, return `None`, or return a box with size `(0,)*D`. My preference would be for the latter option except that it isn't clear where to put origin. Suggestions welcome.

scarlet/source.py
fred3m: I think that since we insist on the convention `(channels, Ny, Nx)` for 3D images while a 2D image will have dimension `(Ny, Nx)` there is no ambiguity here. Since morphology is always 2D, it's consistent with this convention to use `(0,1)`. But I do see how this can be confusing and wouldn't be opposed to `(-2, -1)` for clarity.

scarlet/bbox.py
fred3m: Morphology isn't 3D, so I think that it's ok to have 2D boxes, but this makes me think that there should be a way to "slice" a box to take a 3D box and extract only the 2D portion of it (which can obviously be done by hand using the `from_bounds` method but this might be nice syntactic sugar.

pmelchior: We only use 2D and 3D (so far?...), however, we implicitly cast 2D boxes to 3D. that create a lot of code bulk, which this PR removes. I can think of 4D (Time, Channel, Height, Width) or so, but it's largely about code structure not functionality.

scarlet/bbox.py
pmelchior: This version maintains the signature of the method, yours doesn't. I wanted to make minimal changes.

scarlet/bbox.py
pmelchior: `self` doesn't exist yet. that's a `classmethod`.

scarlet/source.py
pmelchior: It's fully consistent. This method internally creates the box from a 2D image, hence it's 2D. Nobody else touches the bbox, so as long as the author of that piece of code knows how BBox behaves, it's all good.

scarlet/bbox.py
fred3m: In general I'm in favor of minimal API changes but since we just merged a major API change yesterday, making another slight tweak before we've even publicized these changes seems reasonable. As you mention in your previous comment, if we got to ever greater dimensionality like `(time, channel, height, width)` it gets increasingly difficult to make sure that we are passing the appropriate pairs of boundaries.

pmelchior:
Seems to work now, dunno why...

scarlet/bbox.py
pmelchior: I don't think that box needs to be immutable when both `shape` and `origin` are tuples. I'll cache the `stop` values.

fred3m: If they aren't immutable then it's possible for `origin` and `shape` to be overwritten by a user, which would make a cached `stop` incorrect.  So we either need to make boxes immutable (which is what most codes do) or define setters for `origin` and `shape` that will update `stop` if either the `shape` or `origin` changes.

scarlet/bbox.py
herjy: Thanks for clearing this out. 
And yes, I can see how ND data can exist, but it would be conceptually a different regime.

scarlet/bbox.py
pmelchior: That's why I coded it as property

scarlet/bbox.py
pmelchior: Solved it by checking that bounds are ordered (min, max), which would be violated for non-overlapping boxes.

scarlet/source.py
pmelchior: changed to (-2, -1)

herjy: Yup, I missed the 2D nature of the morphologies, fine. I guess my concern was more about generalising to higher dimensions but since we saw earlier that this is not the intended purpose, it's ok. Still (-2,-1) is nice.

scarlet/bbox.py
herjy: I prefer the latter option of keeping this flexibility.

scarlet/bbox.py
fred3m: What does it do if they are not ordered?

pmelchior: shape = 0

fred3m: Ok, so looking at the code if I have one box from `(0,0)` to `(1,1)` and another from `(0,2)` to `(2,3)` it looks like the result of `box1 & box2` would be a new box with origin `(0,2)` and `shape=(1,0)`. This isn't really the overlap, since there is no overlap. In other words the shape should be `shape=(0,0)`. And, as I mentioned above, this leaves us with a decision to make about where to put the origin of an empty box, since the concept of an origin doesn't really make sense here. If we decide that we should return an empty box, then I would want the return to be `Box((0,)*self.D, origin=?)`.

I still don't like the idea of an empty box having an origin unless it's something like `(np.inf, np.inf)` or `None`, since it could be confusing to set `overlap = (box1 & box2)` and then go on performing calculations with `overlap.origin`. So I'm starting to lean more towards returning `None` if there is no overlap or inherit a new `EmptyBox` class from `Box` where `EmptyBox` defines a box with `shape=(0,0)` and `origin=None`.

pmelchior: I disagree with both paragraphs:
* since the logic behind the nD code is that boxes are n intervals, one for each dimension, the logic of your overlap example is correct: there is overlap in the first dimension, but not in the last. this means that the overlap has no volume, and no point is contained. I don't see any relevant arithmetic that can be flawed because the Box has some non-zero shape and an arbitrary origin. In particular, the slice of this box is empty for any array given. That's the function the BBox needs to implement.
* Your suggestion necessitates extra code and checking of corner cases. Mine doesn't.

scarlet/bbox.py
pmelchior: I think @herjy, you refer to the code as is, i.e. with a tuple that is generated on the fly. So do I. I don't think this is where we spend most of our time. If so, I'm happy to precompute it.

scarlet/bbox.py
fred3m: I get your point, in that if we project the boxes onto the first dimension then there would be an overlap, but there is still an inconsistency. For example, use the overlap box described above, namely
```
overlap = Box(shape=(1,0), origin=(0,2))
```
Now calculate `overlap.contains((1,2))`, which results in `True`! Of course had we chosen a different coordinate for the arbitrary (non-overlapping) dimension, for example `origin=(0,0)` then `overlap.contains((0,0))` would return `True` and all other points would return `False`. So no matter where we put the origin, as long as it's a finite number then there will always be one point contained in a non-overallping `Box` that your code returns, and it's an arbitrary point.

I think that we can all agree that no points should be contained in a box with zero volume, right? So at the very least we need to add a check in `contains` that all of the dimensions are non-zero, but preferably if you want to allow for overlapping projections then I would want the origin to be `(0, None)` or `(0,np.inf)` to prevent improper interpretation of the origin.

pmelchior:
```
overlap = scarlet.Box(shape=(1,0), origin=(0,2))
overlap.contains((1,2))
```
yields `False`, as it should.

pmelchior: Same for `overlap.contains(overlap.origin)` or any other point

fred3m: Ah, I didn't notice that you changed that line in `contains` this morning. So while I don't like this behavior, I think that I won't like whatever we come up with because I don't think that there is a good solution in terms of what we call the origin in the non-overlapping dimension. So I'm willing to accept this for now with the caveat that we might need to revisit this later.

pmelchior: Yeah, I fixed that logic bug in `contains` so that everything is consistent with python slicing. For what it's worth, I also don't *like* a dangling `origin`, but I just don't think it's worth extra code.

Issue #120, opened 2019-07-09
Flexible frame configurations (e.g. intersection/union handling)


herjy:
At the moment, multi resolution reconstructs the scene inside the footprint of the high resolution image, while accounting for all the low resolution pixels. This works best if the high resolution frame is fully contained in the low resolution one.
We should leave it as an option for users to reconstruct scenes in the union or overlap of all frames across resolutions.

pmelchior:
It's a more general issue: we assume that the frame of Blend is contained in the data, even if all frames have the same resolution.

We should indeed have the ability to chose the data and blend frames flexibly, which would account for union/intersections. Renamed the issue to clarify that.

pmelchior:
Beyond this issue, we should think about what really constitutes the blend frame. In essence, it's a piece of the sky model. How to segment it is entirely non-trivial, but necessary to run at survey scale.

herjy:
Being able to make models that live in the union of two observations rose a chicken and egg problem of the sort. the union requires that we know the common `frame` that encapsulates 2 `observations`. But the `frame` is defined outside the `observation` class and therefore cannot know the size of the picture that encapsulates both observations. 
What I propose is that we create a class `MultiObservations` that would take as arguments a set of observations and a reference psf and that would return the arguments to set up the `frame`. Then this `frame` could be matched to the `observations` by calling `MultiObservation.match(frame)`. Alternativelly, `Frame_MultiObservations` could inherit from frame, generate the frame and operate the matchings at the same time. This way, the whole 5th block in notebook `tutorials/multiresolution.ipynb` would be replaced by:
```
obs_hst = scarlet.Observation(data_hst, wcs=wcs_hst, psfs=psf_hst, channels=channels_hst)
obs_hsc = scarlet.LowResObservation(data_hsc,  wcs=wcs_hsc, psfs=psf_hsc, channels=channels_hsc)
frame = Frame_MultiObservations([obs_hsc, obs_hst], psf_hst, wcs_hst)
```
This is a small change in the user interface, but it only affects the multi-resolution usecase, and it makes things, I think, clearer from the user's point of view.

pmelchior:
I'm thinking of something similar, but without the need for a special class. Instead of line 3 in your example, I previously suggested a class method:
```
frame = Frame.from_observations(observations, perimeter='intersection')
```
where `perimeter in ['intersection', 'union']`.

[...]

PR #155, opened 2020-01-07
Fix shifts

herjy:
This PR fixes a few bugs that were previously affecting the alignment of multiresolution frames. 
* In particular it preforms a needed axis inversion by inverting the sign of the argument of a shift, which is the correct mathematical way of doing it here, instead of inverting the axis of an array.
* The other source for unwelcome shifts was the inconsistency between even and odd shape paddings of images at different resolutions. 
* Another correction that seems to have no effect was to avoid setting arguments of a complex exponential as an axponent. This should generally be avoided as complex exponentials do not behave in the same way as the exponential function in R.


pmelchior:
Can you check the tests, please

herjy:
shifts were handled inconsistently between the `adam` and the `fix_shifts` branch. Fixed now.

pmelchior:
Please remove the file `docs/hsc_cosmos_35.sca`. It should not be tracked.

pmelchior:
I'll do a few more checks later today, so expect a few comments then.

scarlet/resampling.py
```diff
@@ -1,11 +1,10 @@
 import numpy as np
 
 
-def match_patches(
-    shape_hr, shape_lr, wcs_hr, wcs_lr, isrot=True, perimeter="overlap", psf=False
-):
+def match_patches(shape_hr, shape_lr, wcs_hr, wcs_lr, isrot = True, perimeter  = 'overlap'):
```
pmelchior: Can you please rename `perimeter` to `coverage` and `overlap` to intersection.

pmelchior: This affects several places in this method

scarlet/observation.py
```diff
@@ -434,6 +414,7 @@ def match(self, model_frame):
             model_frame.wcs,
             self.frame.wcs,
             isrot=self.isrot,
+            perimeter = perimeter
```
pmelchior: I don't think this is the way to do it. `perimeter` (or better: `overlap`) needs to affect the construction of the model frame, not the matching of the model frame and the observed frame. The latter really only needs the  intersection plus the width of the diff kernel. This will have to be dealt with in another PR, which also ties in #120. See my code recommendation there.

scarlet/observation.py
```diff
@@ -338,49 +323,34 @@ def sinc_shift(self, imgs, shifts, axes, sum_axis):
...
-            inv_shape,
-            np.array(axes) + len(imgs_shiftfft.shape) - 2,
-        ).image
-
+        # The n-dimensional transform could pose problem for very large images
+        op = fft.Fourier.from_fft(imgs_shiftfft, fft_shape, inv_shape, fft_axes).image
         return op
 
-    def match(self, model_frame):
+    def match(self, model_frame, perimeter = 'union'):
```
pmelchior: See next comment about `perimeter` in `match`

pmelchior:
`LowResObservation.match` seems to have slowed down considerably. I remember taking only 200ms, and it's now around 1s. Do you know the reason? Or am I misremembering?

PR #158, opened 2020-01-23
Warm-start

pmelchior:
Storing `m`, `v`, `vhat` in the parameter allows `adaprox` to pick up the optimization where it stopped. 
This is important if we want to add/change sources based on the current model or its residuals.

When serializing the sources, these extra arrays (each has the same size as the parameter itself) could be removed, but then refitting as in the quickstart tutorial becomes less stable / quick. It should be left to the user to decide of they are kept or not.


pmelchior:
1. I don't think that these variables need to be named like you suggest. These are the names used in all papers that are based on the adam optimizer, including our own. I *should* add suitable docstrings in `Parameter`, though (I thought I had).
2. The old version of the quicktstart guide used the stupid way of refitting the model with two more sources. With the new warm start, you can avoid the freeze entirely because the optimizer uses the current m,v,vhat for the existing parameters, thus giving them a lot more stability. This is why these wild swings in logL are a relic of the past. In addition, I prefer the cleaner per-parameter freezing, which can still be done as you say, over the somewhat opaque per-component freeze. My mentality on special-purpose method like `freeze` is to avoid them as long as their function can be replicated with lower-level primitives (I believe I have been influence by this [video](https://www.youtube.com/watch?v=wf-BqAjZb8M)).
3. To the degree I can tell, the only change to `proxmin` was that we can pass `M`, `V`, `Vhat`. That's not a fundamental change of its capabilities. I increased its version counter, pypi package, and the import on our side. I believe that is sufficient.

fred3m:
1. These aren't user facing, so while I still think it is bad practice it's ok as long as the docs are updated.

2. That makes sense. I'll try to find time to watch the video later but I can by the argument in general.

3. That's all well and good except that pypi doesn't send out notifications that the version has been changed, and neither does github. So anyone using `proxmin` will not know that there has been a feature update. I follow a lot of other packages, like `autograd`, `flutter`, etc. and even projects with small teams create feature branches for everything to notify watchers of changes. Even hot fixes made directly on github, like the one I made yesterday in scarlet, go through a PR to ensure that users watching the repo are notified.

I see that you updated the docs so I'll mark this as reviewed, but please update `proxmin` from a feature branch in the future to keep us all up to date.

PR #159, opened 2020-01-24
Fix convergence

fred3m:
When there were zero weights in an observation (for example masked pixels) a division by zero created an infinite value that caused the model to never converge.

### DIFF
--- a/scarlet/display.py
+++ b/scarlet/display.py
@@ -219,10 +219,11 @@ def show_scene(
         ax = (ax,)
 
     # Mask any pixels with zero weight in all bands
-    mask = np.sum(observation.weights, axis=0) == 0
-    # if there are no masked pixels, do not use a mask
-    if np.all(mask == 0):
-        mask = None
+    if observation is not None:
+        mask = np.sum(observation.weights, axis=0) == 0
+        # if there are no masked pixels, do not use a mask
+        if np.all(mask == 0):
+            mask = None
 
     panel = 0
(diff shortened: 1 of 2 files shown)


scarlet/display.py
```diff
@@ -146,12 +141,10 @@ def img_to_3channel(img, channel_map=None, fill_value=0):
     return rgb
 
 
-def img_to_rgb(img, channel_map=None, fill_value=0, norm=None):
+def img_to_rgb(img, channel_map=None, fill_value=0, norm=None, mask=None):
```
pmelchior: why is `mask` given to this function? it can work happily without as it doesn't affect the conversion from pixels to RGB at all

fred3m: It's a convenience to mask pixels in the function instead of having to do it manually, otherwise we need to add the `dstack` method to `show_scene` and users would have to manually mask the image.

More importantly it will always mask the image when a pixel is masked in every band, which can be useful to know if one does not know that pixels were masked.

scarlet/observation.py
```diff
@@ -165,9 +165,12 @@ def get_loss(self, model):
         # with inverse variance weights: sigma^2 = 1/weight
         # full likelihood is sum over all data samples: pixel in images
         # NOTE: this assumes that all pixels are used in likelihood!
```
pmelchior: please add this change also to `LowResObservation` in lines 610 ff

PR #161, opened 2020-03-03
Deconvolution

fred3m:
Implement new classes and notebooks for deconvolving input images to improve initialization.

### DIFF
--- a/scarlet/display.py
+++ b/scarlet/display.py
@@ -182,6 +182,7 @@ def show_scene(
     show_residual=False,
     label_sources=True,
     figsize=None,
+    linear=True,
 ):
     """Plot all sources to recreate the scence.
     The functions provides a fast way of evaluating the quality of the entire model,
@@ -202,6 +203,9 @@ def show_scene(
     label_sources: bool
         Whether each source is labeled with its numerical index in the source list
     figsize: matplotlib figsize argument
+    linear: bool
+        Whether or not to display the scene in a single line (`True`) or
+        on multiple lines (`False`).
     Returns
     -------
     matplotlib figure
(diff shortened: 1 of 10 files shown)


pmelchior:
Am I correct to assume that this PR supersedes #132? If so, please close it.

fred3m:
I'm fine with closing that PR but I want to save the branch just to have a record of that work. But if you want I can rename it something less opaque than the DM ticket number.

scarlet/display.py
```diff
@@ -271,7 +282,6 @@ def show_scene(
                 ax[panel].text(*center_[::-1], k, color="w")
 
     fig.tight_layout()
-    plt.close()
```
pmelchior: Can you confirm that the notebooks don't show the figures twice without this line?

scarlet/observation.py
```diff
@@ -78,7 +80,16 @@ def match(self, model_frame):
...
+        diff_kernels: array
+            The difference kernels to implement.
+            Typically the calculation of `diff_kernels` is the primary task
+            of this method and this parameter is `None`,
+            however in some instances, for example deconvolution,
+            the difference kernels may have been precalculated.
+            In that case the `match` method is merely responsible for setting
+            the internal `_diff_kernels` attribute and making sure that the
+            internal properties exist that are necessary to convolve an input
+            image with the difference kernels in the `render` method.
```
pmelchior: Your docstrings are way too verbose. I'd state the format expected and that it replaces/prevents the internal diff-kernel computation

scarlet/display.py
fred3m: This is why I had to modify the notebooks. The reason that it was showing twice was the this method returns the figure, which is good because it gives users the option of modifying it. So you'll notice in the docs that I had to add a `plt.show()` method, which displays the current figure. Really any line after the `show_scene` function would prevent the figure from displaying twice, but `plt.show()` is the right way to do it.

PR #162, opened 2020-03-17
Fix rotate

herjy:
That PR is bigger than anticipated, but it is not overly complex. The biggest part of it is the addition of the `from_observations` method to frame, which builds a frame from a set of observations and makes a few decisions as to what resolution and psf to use. Please focus your review on this method as it is the most crucial part. 
Most of other changes have been made to fix bugs in the alignment of observations. I have also built a few functions that make (I think), the code slightly easier to read at places.


scarlet/frame.py
```diff
@@ -105,5 +108,171 @@ def get_sky_coord(self, pixel):
...
+        for c, obs in enumerate(observations):
+            #Concatenate all channels
+            channels = channels + obs.frame.channels
+            #concatenate all pixel sizes
+            pix_tab.append(interpolation.get_pixel_size(interpolation.get_affine(obs.frame.wcs)))
+            h_temp = interpolation.get_pixel_size(interpolation.get_affine(obs.frame.wcs))
+            # Looking for the sharpest and the fatest psf
+            for psf in obs.frame._psfs.image:
+                psf_size = get_psf_size(psf)*h_temp
+                if (fat_psf_size is None) or (psf_size > fat_psf_size):
```
fred3m: small detail, but since this is an `or` condition I would change the order, since `fat_psf_size` will only be true once whereas the second condition might be true multiple times.

scarlet/frame.py
```diff
@@ -105,5 +108,171 @@ def get_sky_coord(self, pixel):
...
+                    fat_psf_size = psf_size
+                if (obs_id is None) or (c == obs_id):
+                    if (target_psf is None) and ((small_psf_size is None) or (psf_size < small_psf_size)):
```
fred3m: see previous comment

scarlet/frame.py
```diff
@@ -105,5 +108,171 @@ def get_sky_coord(self, pixel):
...
+    -------
+        sigma3: `float`
+            radius of the area inside 3 sigma around the center in pixels
+    """
+    # Normalisation by maximum
+    psf_frame = psf/np.max(psf)
+
+    # Pixels in the FWHM set to one, others to 0:
+    psf_frame[psf_frame>0.5] = 1
+    psf_frame[psf_frame<=0.5] = 0
```
fred3m: clever!

herjy:
I think you've never seen that PSF size estimate because it is not super accurate, but since the idea here is just to have a size for a margin to pad, I think it's a safe use of a super simple algo.

fred3m:
I agree that it isn't accurate, especially since PSF are better approximated as a Moffat or multiple Gaussians. But I agree that it should be a good enough estimate for comparing PSF sizes and would be a good initial guess for someone who was trying to actually fit a function to a Gaussian.

scarlet/observation.py
```diff
@@ -157,8 +154,8 @@ def get_loss(self, model):
         """
 
         model_ = self.render(model)
-        images_ = self.images[self.slices]
-        weights_ = self.weights[self.slices]
```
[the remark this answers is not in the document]
herjy: Yes, I think this is the right way. The slices have been introduced by @pmelchior I think, in the 1.0 overhaul (and I hate it syntactically speaking, though I recognise its efficiency). The idea is to have slices of the model frame to be able to cutout/insert the model in the image. This is now handled by render and the loss is computed over the whole image,  not just the model. I have now updated render to handle this correctly and adopted the same form for `LowResObservation`. I have also changed the name form slices  -> model_slices

@pmelchior should probably look at it and validate.

scarlet/observation.py
```diff
@@ -167,24 +164,34 @@ def get_loss(self, model):
...
 class LowResObservation(Observation):
     def __init__(
-        self,
-        images,
-        wcs=None,
-        psfs=None,
-        weights=None,
-        channels=None,
-        padding=3,
+            self,
```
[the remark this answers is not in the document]
herjy: No idea, I think this is my automated formatting that does that. I had it set to pep8, but it does not seem happy with the last changes you made.

scarlet/frame.py
```diff
@@ -105,5 +108,171 @@ def get_sky_coord(self, pixel):
...
+                    fat_psf_size = psf_size
+                if (obs_id is None) or (c == obs_id):
+                    if (target_psf is None) and ((small_psf_size is None) or (psf_size < small_psf_size)):
+                        small_psf_size = psf_size
+                        target_psf_temp = PSF(psf[np.newaxis, :, :])
+                        psf_h = h_temp
+
+        # Find a reference observation. Either provided by obs_id or as the observation with the smallest pixel
+        if obs_id is None:
+            obs_ref = observations[np.where(pix_tab == np.min(pix_tab))[0][0]]
```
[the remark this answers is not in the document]
herjy: The psf's case is handled a few lines above. check line 156. The conditioning of this initialisation depending on the arguments gave me huge headaches, so thanks for checking. 

I tested all case of initialisation and they should all work.

scarlet/frame.py
herjy: Gotcha!

scarlet/frame.py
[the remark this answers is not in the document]
herjy: Also the lines below help understanding what is made of the psf.
It is split because I to identify the reference observation to get its wcs.

scarlet/frame.py
herjy: Nope, because if fat_psf_size is `None`, it does not work.

scarlet/frame.py
herjy: See previous comment

herjy:
In my latest push I just trimmed the diff_kernel to the minimal padding needed. This sacrifices a bit on the runtime as the previous padding was insufficient and partly responsible for some of the artefacts. On the attached plot you see the new timings with a relative trend in M/20 compared to galsim. The trend before this new padding scheme is in orange. The change is not dramatic and affects the large sizes more than the small ones where we are doing slightly better.

Screen Shot 2020-03-27 at 4 31 59 PM

scarlet/observation.py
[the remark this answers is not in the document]
pmelchior: The slices have a purpose! They are needed of the `model_frame` has a different bounding box as the observations. You need to keep them wherever you make comparisons between the model and the data.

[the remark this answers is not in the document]
herjy: Yes, I get that it's neede, but I changed the way this is written. Now render slices the model to the image's frame, not the opposite. And I think that how it should be done

scarlet/observation.py
```diff
@@ -135,9 +134,10 @@ def render(self, model):
             `model` mapped into the observation frame
         """
 
-        image_model = model[self.slices]
         if self._diff_kernels is not None:
-            image_model = self._convolve(image_model)
+            image_model = self._convolve(model)[self.model_slices]
+        else:
+            image_model = model[self.model_slices]
```
herjy: here

scarlet/observation.py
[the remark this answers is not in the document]
pmelchior: No, black would remove them. Doesn't matter.

scarlet/observation.py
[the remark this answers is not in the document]
pmelchior: OK, but where are `weights` sliced? It's not in render (because why would it), and it's not here either.

scarlet/observation.py
pmelchior: If I'm reading this right, that only works if the model frame encompasses all the image pixel, which it would in the case of `coverage='union'` but not for `intersection`, right?

scarlet/observation.py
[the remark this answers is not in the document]
pmelchior: I see what's done here. I think that makes sense to do it in render, but see my caveat there.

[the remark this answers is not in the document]
herjy: But the weights have the same size as the image ergo, they need not be sliced. Now we slice the model to the shape of the image.

scarlet/observation.py
herjy: No, on the contrary. The slices correspond to the size of the frame (I should name them frame_slices). In the tutorials, it works with both union and intersection.

herjy: And it's a case where the union and intersection frame are not identical because the LowRes observation is large than the other.

scarlet/fft.py
```diff
@@ -106,6 +106,10 @@ def _get_fft_shape(img1, img2, padding=3, axes=None, max=False):
     while shape[-1] % 2 != 0:
         shape[-1] += 1
         shape[-1] = fftpack.helper.next_fast_len(shape[-1])
+    if shape2[-2] % 2 == 0:
+        while shape[-2] % 2 != 0:
+            shape[-2] += 1
+            shape[-2] = fftpack.helper.next_fast_len(shape[-1])
```
fred3m, line 112: This check doesn't make sense to me. `if shape2[-2] % 2 == 0` means that `shape2[-2]` is even, so when is the while loop ever executed?

scarlet/frame.py
```diff
@@ -180,99 +185,74 @@ def from_observations(observations, target_psf = None, target_wcs = None, obs_id
...
-        # Shape of the resulting frame
-        ny = (y_fmax - y_fmin + 1).astype(int)
-        nx = (x_fmax - x_fmin + 1).astype(int)
-        frame_shape =(len(channels), ny, nx)
-
-        fbox = Box(frame_shape, origin = (0,np.around(y_fmin).astype(int), np.around(x_fmin).astype(int)))
+                if coverage == 'union':
+                    ref_box = ref_box | Box((obs.frame.C, y_max - y_min + 1, x_max - x_min + 1),
+                                            origin = (0, y_min, x_min))
+                elif coverage == 'union':
```
fred3m, line 210: `if coverage == 'union'` and `elif coverage == 'union'` ensure that the following block will never run

fred3m: You can also probably simplify this to something like
```
new_box = Box((obs.frame.C, y_max - y_min + 1, x_max - x_min + 1), origin = (0, y_min, x_min))
if coverage == 'union':
    ref_box |= new_box
else:
    ref_box &= new_box
```

since you already checked that `coverage` is either `union` or `intersection`.

scarlet/frame.py
```diff
@@ -180,99 +185,74 @@ def from_observations(observations, target_psf = None, target_wcs = None, obs_id
...
-    psf_frame[psf_frame>0.5] = 1
-    psf_frame[psf_frame<=0.5] = 0
-
-    # Area in the FWHM:
-    area = np.sum(psf_frame)
-
-    # Diameter of this area
-    d = 2*(area/np.pi)**0.5
+    def __and__(self, other):
+        """Intersection of two bounding boxes
```
fred3m, line 228: Intersection of two `Frame`s?

scarlet/observation.py
```diff
@@ -64,7 +63,7 @@ def __init__(
         assert (
                 self.weights.shape == self.images.shape
         ), "Weights needs to have same shape as images"
-
+        self.model_images = np.zeros(self.frame.shape)
```
fred3m, line 66: I don't see where this is used. Did I miss something?

scarlet/psf.py
```diff
@@ -253,8 +255,8 @@ def get_loss(self, model):
         relative error.
         """
         model_ = self.render(model)
-        images_ = self.images[self.slices]
-        weights_ = self.weights[self.slices]
+        images_ = self.images[self.slices_for_images]
+        weights_ = self.weights[self.slices_for_images]
```
fred3m, line 259: I'm not sure that I understand the need for this change, having both a `slices` and `slices_for_images` property. I agree that `slices_for_images` is more clear, but in that case why keep `slices` at all?

scarlet/observation.py
herjy: This has been dealt with in commit 75beb74

scarlet/observation.py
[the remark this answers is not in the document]
herjy: In the Observation class the images and weights are now trimmed in the loss if the frame is smaller than the observation's frame. The rendered model is trimmed if the model frame is larger than the observations's frame.

scarlet/frame.py
```diff
@@ -105,5 +111,148 @@ def get_sky_coord(self, pixel):
...
+        else:
+            assert self.channels is not None and other.channels is not None
+            cmin = list(other.channels).index(self.channels[0])
+            cmax = cmin + self.C + 1
+        bounds.append((cmin, cmax))
+        for d in range(self.D - 1):
+            bounds.append(
+                (max(self.start[d + 1], other.start[d + 1]), min(self.stop[d + 1], other.stop[d + 1]))
+            )
+        return Box.from_bounds(*bounds)
```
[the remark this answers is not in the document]
herjy: I would rather not, because then what psf/weights etc do you use in the output? The advantage of this is sollely to nicely handle the channels.

scarlet/fft.py
herjy: There are 2 different shapes: Shape and shape2. This says that if shape2[-2] is even, so should shape[-2]

fred3m: Ah, so it does!

scarlet/frame.py
herjy: Good catch, that's a major mistake!
And thanks for the simplification! That's that kind of stuff that I should have caught by doing a good second pass.

scarlet/frame.py
herjy: Yes, it's essentially the intersection of the boxes along with handling the channels properly

scarlet/observation.py
herjy: Good catch. A remainder from a previous attempt.

scarlet/frame.py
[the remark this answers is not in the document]
herjy: Although that could be a good place for building the diff kernel. The wcs would be that of self. I'll keep that in mind for possible future  PRs

PR #165, opened 2020-04-17
Implement add in place with gradient

fred3m:
This PR removes the use of padding around sources, replacing it with an in place `primitive` method to calculate the gradient, saving >15% runtime for most blends.

### DIFF
--- a/scarlet/bbox.py
+++ b/scarlet/bbox.py
@@ -98,28 +98,6 @@ class Box:
                 return False
         return True
 
-    def slices_for(self, im_or_shape):
-        """Slices for `im_or_shape` to be limited to this bounding box.
-
-        Parameters
-        ----------
-        im_or_shape: array or tuple
-            Array or shape of the array to be sliced
-
-        Returns
-        -------
-        slices for every dimension
-        """
-        if hasattr(im_or_shape, "shape"):
-            shape = im_or_shape.shape
-        else:
-            shape = im_or_shape
-        assert len(shape) == self.D
-
-        im_box = Box(shape)
-        overlap = self & im_box
-        return tuple(slice(overlap.start[d], overlap.stop[d]) for d in range(self.D))
-
     def as_slices(self):
         return tuple([slice(o, o+s) for o,s in zip(self.origin, self.shape)])
 
--- a/scarlet/bbox.py
+++ b/scarlet/bbox.py
@@ -1,4 +1,38 @@
-import numpy as np
+import autograd.numpy as np
+from autograd.extend import defvjp, primitive
+
+
+@primitive
+def _add_slice(destination, source, dest_slices, src_slices):
+    """Add the source to the portion of destination contained in the box
+    """
+    destination[dest_slices] += source[src_slices]
+    return destination
+
+
+def _grad_add_slice_dest(result, destination, source, dest_slices, src_slices):
+    """Gradient for the destination
+
+    Since this is just addition, the upstream gradient is passed through.
+    """
+    return lambda upstream_grad: upstream_grad
+
+
+def _grad_add_slice_source(result, destination, source, dest_slices, src_slices):
+    """Gradient for the source
+
+    This just extracts the portion of the upstream gradient that
+    overlaps with the source
+    """
+    def result(upstream_grad):
+        _result = np.zeros(source.shape, dtype=source.dtype)
+        _result[src_slices] = upstream_grad[dest_slices]
+        return _result
+    return result
+
+
+# Register this function in autograd
+defvjp(_add_slice, _grad_add_slice_dest, _grad_add_slice_source)
 
 
 class Box:
@@ -169,6 +203,32 @@ class Box:
         imbox -= self.origin
         overlap = imbox & subbox
         image[self.slices_for(image)] = sub[overlap.slices_for(sub)]
+        return image
+
+    def add_into(self, image, sub):
+        """Add the sub-image into the image
+
+        This is the same as insert_into except that the sub is added to the
+        overlapping elements
+
+        Parameters
+        ----------
+        image: array
+            Full image
+        sub: array
+            Extracted sub-image
+
+        Returns
+        -------
+        image: array
+        """
+        imbox = Box.from_image(image)
+        subbox = Box.from_image(sub)
+
+        # imbox now in the frame of this bbox (i.e. of box)
+        imbox -= self.origin
+        overlap = imbox & subbox
+        image = _add_slice(image, sub, self.slices_for(image), overlap.slices_for(sub))
         return image
 
     @property
--- a/scarlet/bbox.py
+++ b/scarlet/bbox.py
@@ -1,4 +1,4 @@
-import autograd.numpy as np
+import numpy as np
 
 
 class Box:
--- a/scarlet/bbox.py
+++ b/scarlet/bbox.py
@@ -1,38 +1,4 @@
 import autograd.numpy as np
-from autograd.extend import defvjp, primitive
-
-
-@primitive
-def _add_slice(destination, source, dest_slices, src_slices):
-    """Add the source to the portion of destination contained in the box
-    """
-    destination[dest_slices] += source[src_slices]
-    return destination
-
-
-def _grad_add_slice_dest(result, destination, source, dest_slices, src_slices):
-    """Gradient for the destination
-
-    Since this is just addition, the upstream gradient is passed through.
-    """
-    return lambda upstream_grad: upstream_grad
-
-
-def _grad_add_slice_source(result, destination, source, dest_slices, src_slices):
-    """Gradient for the source
-
-    This just extracts the portion of the upstream gradient that
-    overlaps with the source
-    """
-    def result(upstream_grad):
-        _result = np.zeros(source.shape, dtype=source.dtype)
-        _result[src_slices] = upstream_grad[dest_slices]
-        return _result
-    return result
-
-
-# Register this function in autograd
-defvjp(_add_slice, _grad_add_slice_dest, _grad_add_slice_source)
 
 
 class Box:
@@ -203,32 +169,6 @@ class Box:
         imbox -= self.origin
         overlap = imbox & subbox
         image[self.slices_for(image)] = sub[overlap.slices_for(sub)]
-        return image
-
-    def add_into(self, image, sub):
-        """Add the sub-image into the image
-
-        This is the same as insert_into except that the sub is added to the
-        overlapping elements
-
-        Parameters
-        ----------
-        image: array
-            Full image
-        sub: array
-            Extracted sub-image
-
-        Returns
-        -------
-        image: array
-        """
-        imbox = Box.from_image(image)
-        subbox = Box.from_image(sub)
-
-        # imbox now in the frame of this bbox (i.e. of box)
-        imbox -= self.origin
-        overlap = imbox & subbox
-        image = _add_slice(image, sub, self.slices_for(image), overlap.slices_for(sub))
         return image
 
     @property
--- a/scarlet/bbox.py
+++ b/scarlet/bbox.py
@@ -23,7 +23,7 @@ class Box:
     """
 
     def __init__(self, shape, origin=None):
-        self.shape = tuple(shape)
+        self._shape = tuple(shape)
         if origin is None:
             origin = (0,) * len(shape)
         assert len(origin) == len(shape)
@@ -120,6 +120,9 @@ class Box:
         overlap = self & im_box
         return tuple(slice(overlap.start[d], overlap.stop[d]) for d in range(self.D))
 
+    def as_slices(self):
+        return tuple([slice(o, o+s) for o,s in zip(self.origin, self.shape)])
+
     def extract_from(self, image, sub=None):
         """Extract sub-image described by this bbox from image
 
@@ -138,12 +141,9 @@ class Box:
 
         if sub is None:
             sub = np.zeros(self.shape)
-        subbox = Box.from_image(sub)
 
-        # imbox now in the frame of this bbox (i.e. of box)
-        imbox -= self.origin
-        overlap = imbox & subbox
-        sub[overlap.slices_for(sub)] = image[self.slices_for(image)]
+        im_slices, sub_slices = overlapped_slices(imbox, self)
+        sub[sub_slices] = image[im_slices]
         return sub
 
     def insert_into(self, image, sub):
@@ -163,12 +163,9 @@ class Box:
         image: array
         """
         imbox = Box.from_image(image)
-        subbox = Box.from_image(sub)
 
-        # imbox now in the frame of this bbox (i.e. of box)
-        imbox -= self.origin
-        overlap = imbox & subbox
-        image[self.slices_for(image)] = sub[overlap.slices_for(sub)]
+        im_slices, sub_slices = overlapped_slices(imbox, self)
+        image[im_slices] = sub[sub_slices]
         return image
 
     @property
@@ -188,6 +185,10 @@ class Box:
         """Tuple of stop coordinates
         """
         return tuple(o + s for o, s in zip(self.origin, self.shape))
+
+    @property
+    def shape(self):
+        return self._shape
 
     def __or__(self, other):
         """Union of two bounding boxes
@@ -265,12 +266,46 @@ class Box:
         self.origin = tuple([a + o for a, o in zip(self.origin, offset)])
         return self
 
+    def __add__(self, offset):
+        return self.copy().__iadd__(offset)
+
     def __isub__(self, offset):
         self.origin = tuple([a - o for a, o in zip(self.origin, offset)])
         return self
 
+    def __sub__(self, offset):
+        return self.copy().__isub__(offset)
+
     def __copy__(self):
-        return Box(self.shape, offset=self.offset)
+        return Box(self.shape, origin=self.origin)
+
+    def copy(self):
+        return self.__copy__()
 
     def __eq__(self, other):
         return self.shape == other.shape and self.origin == other.origin
+
+
+def overlapped_slices(bbox1, bbox2):
+    """Slices of bbox1 and bbox2 that overlap
+
+    Parameters
+    ----------
+    bbox1: `~scarlet.bbox.Box`
+    bbox2: `~scarlet.bbox.Box`
+
+    Returns
+    -------
+    slices: tuple of slices
+        The slice of an array bounded by `bbox1` and
+        the slice of an array bounded by `bbox` in the
+        overlapping region.
+    """
+    overlap = bbox1 & bbox2
+    _bbox1 = overlap-bbox1.origin
+    _bbox2 = overlap-bbox2.origin
+    slices = (
+        _bbox1.as_slices(),
+        _bbox2.as_slices(),
+    )
+    return slices
--- a/scarlet/blend.py
+++ b/scarlet/blend.py
@@ -1,10 +1,11 @@
 import numpy.ma as ma
 import autograd.numpy as np
 from autograd import grad
+from autograd.extend import defvjp
 import proxmin
 from functools import partial
 
-from .component import ComponentTree
+from .component import ComponentTree, tree_get_model, grad_tree_get_model
 
 
 class Blend(ComponentTree):
@@ -57,6 +58,12 @@ class Blend(ComponentTree):
         # dynamically call parameters to allow for addition / fixing
         X = self.parameters
         n_params = len(X)
+
+        # We have to declare the function that inserts sources into the blend with autograd
+        # This has to be done each time we fit a blend,
+        # since the number of components => the number of arguments, which must be linked to the
+        # autograd primitive function
+        defvjp(tree_get_model, *([partial(grad_tree_get_model, index=k) for k in range(n_params)]))
 
         # compute the backward gradient tree
         grad_logL = grad(self._loss, tuple(range(n_params)))
--- a/scarlet/blend.py
+++ b/scarlet/blend.py
@@ -108,6 +108,13 @@ class Blend(ComponentTree):
 
         return self
 
+    def get_model(self, *parameters):
+        """Override `ComponentTree.get_model` to use the model frame
+
+        See `~scarlet.ComponentTree.get_model` for more info.
+        """
+        return super().get_model(*parameters, frame=self.model_frame)
+
     def _loss(self, *parameters):
         """Loss function for autograd
 
@@ -137,3 +144,13 @@ class Blend(ComponentTree):
 
         if callback is not None:
             callback(*parameters, it=it)
+
+    @property
+    def bbox(self):
+        """Bounding box of the blend
+
+        Override the bounding box of the `ComponentTree`,
+        which includes the area of sources that extend beyond
+        the model frame boundaries.
+        """
+        return self.model_frame.bbox
--- a/scarlet/blend.py
+++ b/scarlet/blend.py
@@ -121,6 +121,7 @@ class Blend(ComponentTree):
         total_loss = 0
         for observation in self.observations:
             total_loss = total_loss + observation.get_loss(model)
+
         self.loss.append(total_loss._value)
         return total_loss
 
--- a/scarlet/blend.py
+++ b/scarlet/blend.py
@@ -32,7 +32,7 @@ class Blend(ComponentTree):
         observations: a `scarlet.Observation` instance or a list thereof
             Data package(s) to fit
         """
-        ComponentTree.__init__(self, sources, trim=False)
+        ComponentTree.__init__(self, sources)
 
         try:
             iter(observations)
--- a/scarlet/blend.py
+++ b/scarlet/blend.py
@@ -59,12 +59,6 @@ class Blend(ComponentTree):
         X = self.parameters
         n_params = len(X)
 
-        # We have to declare the function that inserts sources into the blend with autograd
-        # This has to be done each time we fit a blend,
-        # since the number of components => the number of arguments, which must be linked to the
-        # autograd primitive function
-        defvjp(_add_models, *([partial(_grad_add_models, index=k) for k in range(n_params)]))
-
         # compute the backward gradient tree
         grad_logL = grad(self._loss, tuple(range(n_params)))
         grad_logP = lambda *X: tuple(
--- a/scarlet/blend.py
+++ b/scarlet/blend.py
@@ -1,11 +1,10 @@
 import numpy.ma as ma
 import autograd.numpy as np
 from autograd import grad
-from autograd.extend import defvjp
 import proxmin
 from functools import partial
 
-from .component import ComponentTree, _add_models, _grad_add_models
+from .component import ComponentTree
 
 
 class Blend(ComponentTree):
--- a/scarlet/blend.py
+++ b/scarlet/blend.py
@@ -5,7 +5,7 @@ from autograd.extend import defvjp
 import proxmin
 from functools import partial
 
-from .component import ComponentTree, tree_get_model, grad_tree_get_model
+from .component import ComponentTree, _add_models, _grad_add_models
 
 
 class Blend(ComponentTree):
@@ -33,7 +33,7 @@ class Blend(ComponentTree):
         observations: a `scarlet.Observation` instance or a list thereof
             Data package(s) to fit
         """
-        ComponentTree.__init__(self, sources)
+        ComponentTree.__init__(self, sources, trim=False)
 
         try:
             iter(observations)
@@ -63,7 +63,7 @@ class Blend(ComponentTree):
         # This has to be done each time we fit a blend,
         # since the number of components => the number of arguments, which must be linked to the
         # autograd primitive function
-        defvjp(tree_get_model, *([partial(grad_tree_get_model, index=k) for k in range(n_params)]))
+        defvjp(_add_models, *([partial(_grad_add_models, index=k) for k in range(n_params)]))
 
         # compute the backward gradient tree
         grad_logL = grad(self._loss, tuple(range(n_params)))
--- a/scarlet/component.py
+++ b/scarlet/component.py
@@ -590,7 +590,10 @@ class ComponentTree:
             (Bands, Height, Width) data cube
         """
         if frame is None:
-            frame = Frame(self.bbox, dtype=self.model_frame.dtype, psfs=self.model_frame.psf)
+            frame = Frame(
+                self.bbox, dtype=self.model_frame.dtype,
+                psfs=self.model_frame.psf, channels=self.model_frame.channels
+            )
         # If this is the model frame then the slices are already cached
         if frame == self.model_frame:
             use_cached = True
--- a/scarlet/component.py
+++ b/scarlet/component.py
@@ -4,6 +4,7 @@ from . import fft
 from . import interpolation
 from .bbox import Box
 import autograd.numpy as np
+from functools import partial
 from autograd.extend import defvjp, primitive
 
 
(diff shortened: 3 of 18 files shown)


scarlet/blend.py
```diff
@@ -58,6 +59,12 @@ def fit(self, max_iter=200, e_rel=1e-3, **alg_kwargs):
         X = self.parameters
         n_params = len(X)
 
+        # We have to declare the function that inserts sources into the blend with autograd
+        # This has to be done each time we fit a blend,
+        # since the number of components => the number of arguments, which must be linked to the
+        # autograd primitive function
+        defvjp(_add_models, *([partial(_grad_add_models, index=k) for k in range(n_params)]))
```
pmelchior, line 66: Shouldn't that be declared in `ComponentTree` so that it's available for trees outside of `Blend`?

scarlet/component.py
```diff
@@ -525,19 +539,36 @@ def get_model(self, *params):
...
+            models.append(model)
+
+            # Get the slices needed to insert the model
+            imbox = Box.from_image(full_model)
+            subbox = Box.from_image(model)
+            imbox -= c.bbox.origin
+            overlap = imbox & subbox
+            slices.append((c.bbox.slices_for(full_model), overlap.slices_for(model)))
+
+        full_model = _add_models(*models, full_model=full_model, slices=slices)
```
pmelchior, line 566: You are calling the function directly here, I think it's correct to also define the gradient primitive here instead of in `Blend`

scarlet/blend.py
fred3m: This is the tricky bit about this whole operation. You have to register a `primitive` with autograd with the number of parameters that will be updated by the function with a unique gradient function for each. So we can't do this until we build the blend and know how many components that it has.

My initial run at this (5135580fd15fd9cf7172f5d50f594c058dd01ade) just updated `Box` with a method that updated the gradient each time a model was added, but that still had an overhead because the `model` was being re-written in each step. This cost us ~10% in runtime and with autograd there is no way to avoid this without really going deep into the code and modifying the VJPNode myself (which I started to look into but after a few hours down that rabbit hole this seemed like an easier solution).

scarlet/component.py
```diff
@@ -356,31 +340,52 @@ def __init__(self, frame, cube, bbox=None):
...
+        return _result
+    return result
 
 
 class ComponentTree:
     """Base class for hierarchical collections of Components.
     """
 
-    def __init__(self, components):
+    def __init__(self, components, trim=True):
```
pmelchior, line 388: Why is the `trim` option needed? It cuts the model to the intersection of all component boxes, but what's the purpose of doing so?

fred3m: This is the part I'm most unsure about. The single `Component` models now return a model that is _not_ projected into the full model but instead is the size of each source. I feel like `ComponentTree` should do the same thing, except if it's a `Blend` and we want to get the full model to compare with the observation. So this was my first thought about an easy way to do this, but I'll spend some time over the weekend thinking about a better way. If you have one I'm happy to hear it.

scarlet/blend.py
pmelchior: I don't question the approach itself, but any `ComponentTree` knows the number of components it holds. Are you worried about cases, where we're inserting a ComponentTree into a ComponentTree?

fred3m: That's a good point, I didn't think about that. You're right, this should be in the `ComponentTree` instead of `Blend`.

fred3m: Wait, I take that back. I do remember why I did it this way and you're right in your last comment. `defvjp` updates a global parameter and I wanted to make sure that if someone uses hierarchical components it doesn't get re-written by the child `ComponentTree`s.

scarlet/component.py
pmelchior: If `trim==True`, `ComponentTree.get_model` returns a cube without telling the downstream customer the proper box for it. That can't be right

scarlet/blend.py
pmelchior: Then it's brittle, in that the same process cannot declare different `Blend` instances and have them behave consistently

fred3m: True, that's why I put it in `fit` and not `init`. The only way I can see this breaking is if someone has multiple threads all running scarlet at the same time, in the same environment. This is why you don't use global variables.

When we switch to JAX if we run into the same issue it would be a good idea to write a PR that allows us to have local `VJP`s that could be defined for a single `grad` call.

fred3m: But I've been following autograd and I don't think that they're accepting PRs with new features, only bug fixes, to encourage people to use and develop jax. So even if we wanted to write something now they wouldn't accept it.

pmelchior: Yeah, I think we both feel where this is headed...

scarlet/bbox.py
```diff
@@ -265,12 +244,46 @@ def __iadd__(self, offset):
...
+        return Box(self.shape, origin=self.origin)
+
+    def copy(self):
+        return self.__copy__()
 
     def __eq__(self, other):
         return self.shape == other.shape and self.origin == other.origin
+
+
+def overlapped_slices(bbox1, bbox2):
```
pmelchior, line 267: Yes! Much nicer!

scarlet/component.py
```diff
@@ -70,64 +75,37 @@ def parameters(self):
...
-                    )
-                )
-            padded_box = Box.from_bounds(*bounds)
-
-            model_box = self.frame
-            overlap = padded_box & model_box
-            overlap -= padded_box.origin  # now in padded frame
-            self.slices = overlap.slices_for(padded_box.shape)
+        self.model_frame = model_frame
+        self.model_frame_slices, self.model_slices = overlapped_slices(model_frame, self.bbox)
```
pmelchior, line 108: Really: that's it? I worked hard for the code that you now replace with merely 2 lines...

scarlet/component.py
```diff
@@ -141,6 +119,45 @@ def check_parameters(self):
                 msg = "Component {} Parameter {} is not finite:\n{}".format(self, k, p)
                 raise ArithmeticError(msg)
 
+    def project(self, model=None, frame=None):
+        """Project a model into a frame
+
+
+        Parameters
+        ----------
+        model: array
```
pmelchior, line 128: What's the use case for specified model?

pmelchior: I'm not a fan of the term "project" here either. If this is always the component's model, I'd prefer e.g. `Component.model_for(frame)`, but I think I miss a use case here

scarlet/component.py
```diff
@@ -13,18 +18,18 @@ class Component(ABC):
 
     Parameters
     ----------
-    frame: `~scarlet.Frame`
-        The spectral and spatial characteristics of this component.
-    parameters: list of `~scarlet.Parameter`
+    model_frame: `~scarlet.Frame`
+        The spectral and spatial characteristics of the model.
     bbox: `~scarlet.Box`
         Hyper-spectral bounding box
```
pmelchior, line 24: I'd say "Hyper-spectral bounding box of this component" to make it clear.

scarlet/component.py
```diff
@@ -149,24 +166,24 @@ class FactorizedComponent(Component):
 
     Parameters
     ----------
-    frame: `~scarlet.Frame`
-        The spectral and spatial characteristics of this component.
+    model_frame: `~scarlet.Frame`
+        The spectral and spatial characteristics of the full model.
+    bbox: `~scarlet.Box`
+        Hyper-spectral bounding box
```
pmelchior, line 172: same as above

scarlet/component.py
```diff
@@ -221,31 +242,18 @@ def get_model(self, *parameters):
...
-            padded = np.pad(
-                morph, self.pad_width[1:], mode="constant", constant_values=0
-            )
-            return padded[self.slices[1:]]
-        return morph
+        morph = self._shift_morph(shift, morph)
+        model = sed[:, None, None] * morph[None, :, :]
+        # project the model into frame (if necessary)
+        if frame is not None:
+            model = self.project(model, frame)
```
pmelchior, line 255: Oh, I see. Have you checked if this is faster than expanding `morph` to full `model_frame` and then doing the outer product with `sed`? Slicing the full model is an operation that isn't aligned well in memory.

scarlet/component.py
```diff
@@ -272,42 +280,47 @@ class FunctionComponent(FactorizedComponent):
 
     Parameters
     ----------
-    frame: `~scarlet.Frame`
-        The spectral and spatial characteristics of this component.
+    model_frame: `~scarlet.Frame`
+        The spectral and spatial characteristics of the full model.
+    bbox: `~scarlet.Box`
+        Hyper-spectral bounding box
```
pmelchior, line 286: same

scarlet/component.py
```diff
@@ -350,30 +365,55 @@ class CubeComponent(Component):
...
-            return padded[self.slices]
-        return cube
+
+@primitive
+def _add_models(*models, full_model, slices):
+    """Insert the models into the full model
+    """
+    for i in range(len(models)):
+        if hasattr(models[i], "_value"):
+            full_model[slices[i][0]] += models[i][slices[i][1]]._value
```
pmelchior, line 395: The definition of the tuple `slices` needs to be explained in the docstrings, in particular their order.

pmelchior: Or at least as an inline comment for us

scarlet/component.py
```diff
@@ -525,32 +585,60 @@ def get_model(self, *params):
...
-        if len(params):
-            i = 0
-            for k, c in enumerate(self.components):
+
+        # We have to declare the function that inserts sources
+        # into the blend with autograd.
+        # This has to be done each time we fit a blend,
+        # since the number of components => the number of arguments,
+        # which must be linked to the autograd primitive function
+        defvjp(_add_models, *([partial(_grad_add_models, index=k) for k in range(len(self.components))]))
```
pmelchior, line 594: Still not a fan of this idea because it sets a global number of parameters and prevent correct fitting with multiple concurrent instances of `Blend` in a single thread. Nothing we can do about it now, but I suggest to add a comment here than one can avoid this and revert to the original mode.

pmelchior: Also, for clarity, I'd move the entire block of comments and `defvjp` to below the `_add_models` call.

scarlet/display.py
```diff
@@ -237,45 +240,65 @@ def show_scene(
...
+        tree = ComponentTree(sources)
+    if observation is None:
+        model = tree.get_model()
+        extent = get_extent(tree.bbox)
+    else:
+        model = tree.get_model(frame=observation.frame)
+        extent = get_extent(observation.frame)
+
+    if show_model:
+        ax[panel].imshow(img_to_rgb(model, norm=norm, channel_map=channel_map), extent=extent)
```
pmelchior, line 256: setting the extent is fine, but we still need to set `origin='lower'` for consistency.

scarlet/display.py
```diff
@@ -285,15 +308,24 @@ def show_scene(
     return fig
 
 
+def get_extent(bbox):
+    extent = np.array([bbox.start[-1], bbox.stop[-1], bbox.start[-2], bbox.stop[-2]], dtype=float) - 0.5
```
pmelchior, line 312: I don't think the -0.5 is necessary or even plausible

scarlet/display.py
```diff
@@ -321,74 +355,76 @@ def show_sources(
...
             seds = [model.sum(axis=(1, 2))]
-        src.set_frame(frame_)
-        ax[k][panel].imshow(img_to_rgb(model, norm=norm, channel_map=channel_map, mask=model.sum(axis=0)==0))
-        ax[k][panel].set_title("Model Source {}".format(k))
-        if center is not None:
-            ax[k][panel].plot(*center_[::-1], "wx", mew=1, ms=10)
 
-        if show_rendered:
+        if use_mask:
+            mask = model.sum(axis=0) == 0
```
pmelchior, line 380: no! models are not masked. It's visual clutter and conceptually improper to define the mask this way (what if we *want* the model to be zero over some range). I suggest that we set masking consistently in all panels, i.e. we set the observation mask in all panels if `use_mask is True`.

scarlet/display.py
```diff
@@ -321,74 +355,76 @@ def show_sources(
...
             panel += 1
-            model = src.get_model()
-            model = observation.render(model)
-            ax[k][panel].imshow(img_to_rgb(model, norm=norm, channel_map=channel_map))
+
+        if show_rendered:
+            # Center and show the rendered model
+            model = src.project(frame=rendered_box)
+            model = observation.render(model, in_frame=False)
+            ax[k][panel].imshow(img_to_rgb(model, norm=norm, channel_map=channel_map), extent=extent)
```
pmelchior, line 401: but then all panels need to use masks!

scarlet/display.py
```diff
@@ -321,74 +355,76 @@ def show_sources(
...
+                ax[k][panel].plot(*center, "wx", mew=1, ms=10)
             panel += 1
-            ax[k][panel].imshow(
-                img_to_rgb(observation.images, norm=norm, channel_map=channel_map)
-            )
+
+        if show_observed:
+            # Center the observation on the source and display it
+            _images = observation._project(rendered_box)
+            ax[k][panel].imshow(img_to_rgb(_images, norm=norm, channel_map=channel_map), extent=extent)
```
pmelchior, line 413: same

scarlet/display.py
pmelchior: Actually, it should be an argument `mask=None` like in `show_scene` above.

scarlet/frame.py
```diff
@@ -45,11 +45,12 @@ def __init__(
                 self._psfs = PSF(psfs)
 
         assert channels is None or len(channels) == self.shape[0]
+        if channels is None:
+            logger.warning(
+                "No channels specified. Possible, but multi-observation processing will likely fail!")
```
pmelchior, line 50: I don't like this: if it is likely to fail, we should disallow it. Is there an expert use case for `channels=None`?

scarlet/observation.py
```diff
@@ -1,10 +1,101 @@
...
 from . import interpolation
 from . import fft
 from . import resampling
 
+from .bbox import Box, overlapped_slices
+
+from scarlet.operators_pybind11 import apply_filter
+
+
+def get_filter_coords(filter_values, center=None):
```
pmelchior, line 14: I suggest to move this method and the next to `interpolation.py`, where we have similar functions

scarlet/observation.py
```diff
@@ -188,6 +315,31 @@ def get_LowRes(self):
                                  wcs=self.frame.wcs,
                                  channels=self.frame.channels)
 
+    def _project(self, frame, images=None):
```
pmelchior, line 318: that method has another signature as previous ones, I suggest to use this one here for all of them:
`project(self, frame, model=None)` or `..., image=None)`.
Still not a fan of "project"...

scarlet/observation.py
```diff
@@ -117,33 +214,59 @@ def match(self, model_frame, diff_kernels=None):
...
+        if self.convolution == "real":
+            result = _convolve(model, self._diff_kernels.image, self._convolution_slices)
+        elif self.convolution == "fft":
+            result = fft.convolve(fft.Fourier(model), self._diff_kernels, axes=(1, 2)).image
+        else:
+            raise ValueError("`convolution` must be either 'real' or 'fft', got {}".format(self.convolution))
+        return result
 
-    def render(self, model):
+    def render(self, model, in_frame=True):
```
pmelchior, line 246: how can the model not be contained in the observation frame? When would that be useful?

pmelchior: At the very least, this will be a problem for `LowResObservation`

scarlet/psf.py
```diff
@@ -254,8 +253,8 @@ def get_loss(self, model):
         relative error.
         """
         model_ = self.render(model)
-        images_ = self.images[self.slices_for_images]
-        weights_ = self.weights[self.slices_for_images]
+        images_ = self.images#[self.slices_for_images]
```
pmelchior, line 256: remove the commented part, please

scarlet/source.py
```diff
@@ -1,6 +1,10 @@
-from .constraint import *
-from .component import *
-from .bbox import *
+from functools import partial
+
+from .constraint import PositivityConstraint, MonotonicityConstraint, SymmetryConstraint
+from .constraint import NormalizationConstraint, ConstraintChain, CenterOnConstraint
+from .parameter import Parameter, relative_step
+from .component import ComponentTree, FunctionComponent, FactorizedComponent
```
pmelchior, line 6: Is there a reason for this verbosity?

scarlet/component.py
fred3m: Sure, I can change the method name. How about `model_to_frame` to make it clear what the method does. It has to take model as an argument because of autograd. BTW: jax would be the same.

scarlet/component.py
fred3m: During optimization `frame` is `None`, so this line doesn't execute anytime we're crunched for speed. This is more for visualizations.

scarlet/component.py
fred3m: Do you mean add this block to just _above_ the `_add_models` call? We have to declare it to `autograd` before we call it.

As for your first comment, I agree that this isn't ideal and I'll add a comment. But the reason I put it here is because we'll only be optimizing one blend at a time, so while the global variable sucks (and jax unfortunately does the same thing) we should never get a conflict, even with multiple blends.

scarlet/display.py
fred3m: I don't think it matters, since I explicitly set `x_lim` and `y_lim` based on `extent`, so it is always consistent.

scarlet/display.py
fred3m: This is needed when we set extent because of the way that matplotlib sets the grid with extent. Without the `0.5` the x's that mark the centers are located in the corner of a pixel instead of in the center. This is why I was against this change in matplotlib a few years ago.

scarlet/display.py
fred3m: This is something that I specifically need, so I included it here as an option (but no longer the default). When I'm investigating how a particular change is affecting the models I want to know how big the full extent of the model is, if it runs up against the edge of the bounding box, if it is weird or lopsided, etc. Otherwise I have to write my own method in an extension module to do the exact same thing except pass this one option, so I'd rather just keep it here. So I differentiate it here because this `mask` is not the same as the mask-plane `mask` that marks where there is no data to constrain the model.

scarlet/display.py
fred3m: see above, this is a different mask that I use for troubleshooting.

scarlet/frame.py
fred3m: I'm fine with that. I did make it so that `bbox` can no longer be `None` because the code before would have actually failed if sources didn't have them. In this case it won't fail with no `channels` unless the user has mutliple observations with non-overlapping bands , so I left open the possibility of creating a `Frame` without channels. But I'm fine to also make that field required.

scarlet/observation.py
fred3m: This is actually used in several places, most notably in `display.show_sources`. Passing `in_frame=False` allows us to convolve an image (any image) in the model frame to the observed seeing, for example when we just call `src.get_model()` we no longer get the full model but instead receive the stamp that is contained in `src.bbox`. So we don't want `render` to slice that model, we just want to convolve it with the difference kernel and return the result.

scarlet/observation.py
fred3m: I used `images` instead of `model` because unlike `Component._project`, which projects the model into a `Frame`, this projects an image (by default `observation.images`) into a `Frame`, where `images` must be located in `observation.frame`. Originally I didn't have the `images` argument at all but needed to use this to project `observation.weights` as well as `observation.images`.

scarlet/source.py
fred3m: It is highly recommended not to use `*` imports in python. At worst it makes code less readable because it is difficult to tell where a given class/method was declared, and at worst can create weird import loops.

fred3m: For example see https://www.geeksforgeeks.org/why-import-star-in-python-is-a-bad-idea/

scarlet/display.py
pmelchior: You might be right but then it's a side effect. The `origin` argument is a swith between visualizing a matrix or an image. For reasons God only knows mpl defaults to matrix for a method that is designed to visualize an image.

scarlet/display.py
pmelchior: OK, I see. But the need for `origin` remains: https://matplotlib.org/3.1.1/tutorials/intermediate/imshow_extent.html

scarlet/display.py
pmelchior: The full extent of the model is its bbox. Checking for zero value is a check that assumes a particular constraint, i.e. very special purpose. I suggest that you change in favor of specifying the mask explicitly and then provide `mask = model.sum(axis=0) == 0` if you need that for something. This way the code is much more generally useful.

scarlet/display.py
pmelchior: yes, and you still can. but most people will apply only observational masks and expect consistency.

scarlet/display.py
pmelchior: Actually, I still don't understand what you would be using the zero-mask for. I get it for PSFs, but not for regular source models.

scarlet/frame.py
pmelchior: It would certainly be best practice to set the channel labels. But I've used multi-observation cases without that, so I don't think the statement is strictly true that it will likely fail.

I see these options to be valid: 
1. require channels to be set
2. check that multiple observations work without setting channels for any of them

My problem is that the warning itself isn't helpful. It either works or it doesn't. We leave the user to guess.

scarlet/observation.py
pmelchior: I was referring to the order of the arguments. `frame` is required in all cases, but the secondary argument (either `model` or `images`) is not. This is why I'd prefer this signature here of the earlier incarnations.

scarlet/source.py
pmelchior: fine

scarlet/observation.py
pmelchior: This signature is inherited by other observations, for now only `LowRes`, where this strategy may not work. The issue is modularity. Your code assumes something about the kind of `model` that is fed to this method (in particular, that it could come from a boxed source). But the `match` method only matches the full model frame to the observed frame, source boxes are not known or respected.

You've created a secondary mechanism for source-in-box convolution. I actually don't even understand how that works with FFT diff-kernels which have a determined image shape, namely that of the observed image. 
e.g. https://github.com/pmelchior/scarlet/pull/165/files#diff-160912423f14aa02a64f6b86c73d3feaR400
`model` has box shape, but the convolution expects `model_frame shape`. I'm puzzled...

For consistency, that mode would need to be supported for all kinds of observations, only for a minor performance benefit in plotting. I suggest that `display.show_sources` generated full-frame models for the `show_render` option, since `render` is only guaranteed to work with model_frame`, and to remove the `in_frame` option.

scarlet/display.py
fred3m: Oh, I see, good catch. In all of my notebooks I always define `origin="lower"` globally when I import matplotlib (notice in the docs, for example https://github.com/pmelchior/scarlet/blob/master/docs/0-quickstart.ipynb?short_path=3c6e710#L27 we do the same). So I'll make this change to make it consistent.

scarlet/display.py
fred3m: I agree that this mask should be passed in the parameters as opposed to enforcing a the mask for a particular configuration of a model (even if it is the most commonly used model), so I'll update that. And I agree that for the PSF convolved model we want to see which pixels were masked in the observation, which is why we use that mask consistently in `show_scene`.

But when I look at a particular source I want to know how the underlying model behaved. Note that we don't use sparsity, so any zero pixels in an objects `2^n` x `2^n` box really have no flux, whereas pixels with even a small amount of flux (`10^-large number`) will be black and not masked. So it lets me see where the model wants to put power and helps me understand how/why it is doing that (for example because there was a neighbor there). This effect gets blurred out with the PSF, plus the PSF convolution makes this particular mask go away. So it is something that I use for a specific purpose, but use often enough that I don't want to maintain a separate function on my own (I already have to do that with source initialization).

scarlet/display.py
fred3m: See previous post. I changed the boolean `use_mask=False` to `model_mask=None`, which now allows the user to specify a mask for the model. It doesn't make sense to apply the same mask to the rendered image because we would have to create a new "convolved mask" in the rendered space to make this consistent, which would probably be more confusing.

scarlet/frame.py
fred3m: Ok, at first I didn't understand how this ever worked and after looking into this I don't think that it worked as expected. This is the code before Remy's last PR to match the model frame to an observation frame: https://github.com/pmelchior/scarlet/blob/1b6635155bbdd38457819699cd5e9dbb128fd394/scarlet/observation.py#L91-L102 .

So it "worked" when channels was `None` in both the model and observation frame by implicitly assuming that they overlapped on the first channel. So if, for example, a model was initialized with 5 bands and an observation overlapped on bands 4 and 5, then when calling `Observation.get_loss` it would be updating the wrong bands in the model.

The above code block changed in Remy's PR when he modified the `Frame` class to use a different `__and__` method, which still requires channels but only works when comparing two frames. It provided inconsistent behavior if it was comparing a `Frame` and a `Box`, which is undesirable, and is incompatible with the simplified slicing algorithm.

So I don't _think_ that it ever worked with channels that didn't overlap unless you initialized `Frame` with a `Box` that specifically set the origin of the channel dimension correctly, and that only worked because the model PSF is the same in all bands.

So the question is do we _ever_ allow users to not set the channels, knowing that it can't work in multiband mode and might lead to unexpected results? I am leaning toward "no" unless either of you can think of a good reason why we should allow this.

scarlet/observation.py
fred3m: Ah, I see. Yes, I agree that your suggestion is better.

scarlet/observation.py
fred3m: So I looked into it and there a few things going on here. To make it clear, I've bulleted actions to be taken in my comments below.

The first issue is an oversight on my part when I created the `fft` module last spring. I don't think that I appreciated how expensive padding is computationally in numpy, since it seems like a "trivial" operation (more on this below). So it looks like when I re-wrote `Observation.match` to use `fft.convolve` to create the difference kernel it no longer created them in the `model_frame`. My thinking was (probably) that padding takes a trivial amount of time and in general we won't have two images with a matching shape in k-space, so it seemed ok to create the diff_kernel in PSF space and pad both the diff kernel and the model in real time (since we already had to pad the model anyway). But numpy must be doing something weird under the hood, for example

```
import numpy as np

def pad(arr, newshape):
    newshape = np.asarray(newshape)
    currshape = np.array(arr.shape)
    dS = newshape - currshape
    startind = (dS + 1) // 2
    endind = dS - startind
    pad_width = list(zip(startind, endind))
    return np.pad(arr, pad_width)

def pad2(arr, newshape):
    newshape = np.asarray(newshape)
    currshape = np.array(arr.shape)
    dS = newshape - currshape
    startind = (dS + 1) // 2
    endind = dS - startind
    pad_width = list(zip(startind, endind))
    
    result = np.zeros(newshape, dtype=arr.dtype)
    slices = tuple([slice(start, -end) for start, end in pad_width])
    result[slices] = arr
    return result

arr = np.arange(100).reshape(10, 10)
new_shape = (10000, 10000)

%timeit t1 = pad(arr, new_shape)
%timeit t2 = pad2(arr, new_shape)
```

prints

```
343 ms ± 4.21 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
65 µs ± 1.59 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

which is _much_ more expensive than I would have predicted using the strictly numpy code. I probably expected something more like the `pad2` code I wrote above. Most likely _this_ is why our DFT convolutions are so slow, not the back propagation through the FFTs.

* So I suggest that I re-write https://github.com/pmelchior/scarlet/blob/master/scarlet/fft.py#L40 using the above scheme (which will require writing a direct gradient for autograd).

* It seems outside the scope of this ticket to do more, but we should probably open another issue to create the `diff_kernel` in the model frame again because it will save time and can be done in the existing code base. If you really want I can include it in this ticket as well, but be aware that it might change the behavior of your existing notebooks (at least it will for mine since I've been making use of this).

So now you know why it works to convolve with whatever image we like in render, and I do like that flexibility. I don't like putting this in `display` for two reasons:

1. Right now we are always trimming the model in `render` and it isn't clear to the user this is what's happening. It took me a while to figure out that was going on and I know the codebase.

2. I agree with your point about this not working with all observations, in particular `LowResObservation`, but to me this is all the more reason to put this in display. Only the `Observation` knows whether or not it is valid to project the rendered image into to a given frame and putting this function in `display` puts it on the user to know this.

So I suggest the following:

* I rename the boolean `in_frame` argument to `frame`, where a `Frame` to project the observations into can be passed to the method. If `frame` is `None` or is the same as the `model_frame` then it will slice the model appropriately as it does by default now. However if a different `Frame` is passed to observation, then just like `model_for_frame`, `Observation` will have to use a different convolution (since the proposed changes above will by default pre-pad the difference kernel for the model frame) to project the rendered version of the input image back into the input frame. However, if a `frame` is passed to `LowResObservation` then it will throw a `NotImplementedError` until a later date when we may want to implement a more complicated projection algorithm.

scarlet/display.py
pmelchior: It might have some diagnostic value (although I find pixels around zero rarely diagnostically useful), but certainly not enough to justified it being baked into code as it is.

fred3m: Even the current version pushed today?

scarlet/frame.py
pmelchior: I agree with you, we should disallow it. If a pattern is so prone to failure (as it is because even we can't do it correctly), it should be banned. This might explain the color shifts we can see [here](https://pmelchior.github.io/scarlet/tutorials/multiresolution.html#View-Full-Model).

scarlet/display.py
pmelchior: It's fine now

scarlet/observation.py
pmelchior: Wow: that must be record length for a code review comment!

We should move the conversation about the `in_frame` argument offline because that's a bigger can of worms. I do like the suggestion of implementing the FFT padding directly with the gradients, which is what this PR is supposedly about.

pmelchior: And I see: you've already done it. Any noticeable performance gains?

fred3m: Unfortunately not really, maybe a percent or less on a smaller blend. I imagine that for a bigger blend this might become more substantial but for smaller ones this is mostly in the noise.

scarlet/frame.py
fred3m: I don't think that is causing those color shifts because that notebook defines the channels and they are matched properly. But I'll see what this notebook looks like on my branch tomorrow.

fred3m: Color shifts are still there, so this didn't fix that.

scarlet/observation.py
```diff
@@ -131,19 +189,20 @@ def render(self, model):
         ----------
         model: array
             The model from `Blend`
+        in_frame: bool
+            Whether or not the model must be contained in the observation frame
```
pmelchior, line 193: please remove

scarlet/frame.py
pmelchior: I wrote the previous channel matching code. And it did work *under assumptions*, namely that someone construction the model frame properly, i.e. with a concatenation of the channels. I think it is better to be explicit about it even if that causes a minor hassle for the single-observation case.

I'm ok with it now, but it'll surely come up again when we debate the general `Observation` API and background mechanics.

PR #166, opened 2020-04-28
MuSCADeTifying

herjy:
Scarlet, meet wavelets. 
This PR introduces starlets to scarlet (don't mess up that one). This includes:
* A `Starlet` class allows to do transforms and inverse transforms of array. 
* A `StarletSource` that initialises sources as starlet coefficients with an L0 constrain.
* A Notebook that treats the quickstart example with the `StarletSource` model.

I am amazed by how easy it was to plug the starlet model in scarlet. So far, the L0 constrain is badly imposed since it would require some weighting of the different levels, but even with a crude L0, the results are definitely not that bad! You'll see mostly noise features that come from this wrong weighting of the L0, but sources don't seem to try to steel flux from their neighbours.


scarlet/source.py
```diff
@@ -395,6 +396,96 @@ def __init__(self, frame, sky_coord, observations):
...
+            name="sed",
+            step=partial(relative_step, factor=1e-2),
+            constraint=PositivityConstraint(),
+        )
+        # Threshold in units of noise
+        thresh *= np.sum(sed*noise)
+
+        # Starlet transform
+        transform = Starlet(morph)
+        morph = transform.starlet
```
pmelchior, line 466: It would be good to describe the format/dimensions of this array, It's hard to follow the slicing of the thresholds without it

scarlet/source.py
```diff
@@ -395,6 +396,96 @@ def __init__(self, frame, sky_coord, observations):
...
+        # wavelet-scale norm
+        wavelet_norm = transform.norm
+        #One threshold per wavelet scale: thresh*norm
+        thresh_array = np.zeros(morph.shape) + thresh
+        thresh_array[-3:] = thresh_array[-3:] * np.array([wavelet_norm])[..., np.newaxis, np.newaxis]
+        # We don't threshold the last scale
+        thresh_array[...,-1,:,:] = np.inf
+
+        morph_constraint = L0Constraint(thresh_array)
+        morph = Parameter(morph, name="morph", step=1e-2, constraint=morph_constraint)
```
pmelchior, line 476: What's the amplitude of the morph coefficients? I wonder if `step=1e-2` is appropriate

scarlet/source.py
```diff
@@ -395,6 +396,96 @@ def __init__(self, frame, sky_coord, observations):
     def _psf_wrapper(self, *parameters):
         return self.frame.psf.__call__(*parameters, bbox=self.bbox)[0]
 
+class WaveletSource(FunctionComponent):
+    """Source intialized with a single pixel
```
pmelchior, line 400: These docstrings aren't updated. Please explain what the class does

scarlet/source.py
```diff
@@ -395,6 +396,96 @@ def __init__(self, frame, sky_coord, observations):
...
+        super().__init__(frame, sed, morph, self._iuwt, bbox=bbox)
+
+    @property
+    def center(self):
+        if len(self.parameters) == 3:
+            return self.pixel_center + self.shift
+        else:
+            return self.pixel_center
+
+    def _iuwt(self, param):
```
pmelchior, line 487: The class is hidden, but please leave a comment for us to explain what it does

scarlet/wavelet.py
```diff
@@ -0,0 +1,300 @@
...
+        The transform is performed by convolving the image by a seed starlet: the transform of an all-zero
+        image with its central pixel set to one. This requires 2-fold padding of the image and an odd pad
+        shape. The fft of the seed starlet is cached so that it can be reused in the transform of other
+        images that have the same shape.
+    """
+    def __init__(self, image, lvl = None, starlet = None, direct = False):
+        """ Initialise the Starlet object
+
+        Paramters
+        image: Fourier object
```
pmelchior, line 21: This type seems to be an ordinary array, not an instance of `Fourier`

scarlet/wavelet.py
```diff
@@ -0,0 +1,300 @@
...
+            image to transform
+        lvl: int
+            number of starlet levels to use in the decomposition
+        starlet: array
+            Starlet transform of an array
```
pmelchior, line 26: the transform of *which* array? I'm guessing of `image`...

scarlet/wavelet.py
```diff
@@ -0,0 +1,300 @@
...
+
+    @property
+    def norm(self):
+        """The norm of the seed wavelet in each wavelet level (not in coarse wavelet)"""
+        return self._norm
+
+
+    @property
+    def starlet(self):
+        """The real space image"""
```
pmelchior, line 73: I doubt that this docstring is correct

pmelchior: Also, I dislike that the method name is a lower-case version of the class name. This interferes with the convention of using lower-case names for instances of that class, e.g. `starlet = Starlet()`. Can you pick a suitable name that also clarifies what the method will return.

scarlet/wavelet.py
```diff
@@ -0,0 +1,300 @@
...
+        ----------
+        starlet: array
+            The starlet transform of an array to inverse transform.
+        shape: tuple
+            the expected shape of the untransformed image without padding
+
+        Returns
+        -------
+        Starlet: Starlet object
+            the starlet object initialised with the image that corresponds to the inverse transfform of `starlet`
```
pmelchior, line 96: typo: `transfform`

scarlet/wavelet.py
```diff
@@ -0,0 +1,300 @@
...
+            for star in starlet:
+                rec.append(fft._centered(iuwt(star), shape[-2:]))
+            return Starlet(np.array(rec), starlet = starlet)
+
+        return Starlet(fft._centered(iuwt(starlet), shape), starlet = starlet)
+
+
+
+    def transform(self):
+        """ Performs the wavelet transform of an image by comvolution with the seed wavelet and caches the seed
```
pmelchior, line 118: typo: `comvolution`. I'd also expand the docstring. For instance, it's not clear what image the transform acts on because the method just has `self` as argument.

The caching of the seed should explained similarly in the expanded docstring

pmelchior: Same for the return value. Please use the syntax
```
Returns
-------
...
```

scarlet/source.py
```diff
@@ -395,6 +396,96 @@ def __init__(self, frame, sky_coord, observations):
     def _psf_wrapper(self, *parameters):
         return self.frame.psf.__call__(*parameters, bbox=self.bbox)[0]
 
+class WaveletSource(FunctionComponent):
```
pmelchior, line 399: I'd call it `StarletSource` for clarity

scarlet/wavelet.py
```diff
@@ -0,0 +1,300 @@
...
+        return fft._centered(self._starlet,
+                             [self._image_shape[0], self._starlet_shape[-3], *self._image_shape[-2:]])
+
+    @property
+    def shape(self):
+        """The shape of the real space image"""
+        return self._image.shape
+
+    @staticmethod
+    def from_starlet(starlet, shape = None, starlet_shape = None):
```
pmelchior, line 83: I'm not sure how necessary this method is. It's only useful if you have some starlet coefficients lying around ... And as far as I can tell it's only every used the tests and in the docs. In both cases, it's classed to check for the accuracy of the starlet representation of the original image, i.e. on the inverse transform of the class instance itself, so the coefficients don't need to be provided. I think this would be better done by redefining `@property def image` to return that representation instead of the original image.

pmelchior:
The notebook contains images, so it's 600kB. Please empty the outputs before committing the next fixes. We can still get rid of the committed bulk if we squash and merge.

scarlet/source.py
herjy: I think it's fine. Coefficients are fractions of the pixel values, fraction ranging from 0.8 to 0.01 for images with a 100 ish pixels.

scarlet/wavelet.py
herjy: Yup, old implementation, sorry

scarlet/wavelet.py
herjy: No, because Starlet objects can be instanciated with their transform, so as an argument, it can never be the transform of `image`.

scarlet/wavelet.py
herjy: `coefficients` it is

scarlet/wavelet.py
herjy: Done, please check

scarlet/source.py
herjy: Done

scarlet/wavelet.py
herjy: This method is here to create Starlet objects from a set of coefficients even when no image exist for instance after thresholding coefficients. I want to keep this, because I see wavelets better used as part of a constraint more than as a model. This would require the explicit transform and inverse. Also, this is consistent with the way Fourier transform and inverse is implemented.

scarlet/source.py
pmelchior: That's quite a large variation. Right now the small coefficients have step size of order unity per iteration. Have you tried with a relative step size? https://github.com/pmelchior/scarlet/blob/master/scarlet/parameter.py#L109
By default this is 10% of the mean along axis=0. This could easily be modified, but I found that much more suitable unless there's a known absolute scale of the parameter

scarlet/wavelet.py
pmelchior: I'm not sure I understand your logic. It looks like if `starlet` is not set, the init will compute it from `image`. If it is set, init won't compute it (fine), but then it should still be the transform of `image`.

scarlet/source.py
herjy: I'll try that, but there aree other issues in my implementation I think at the moment that have a larger impact.

scarlet/wavelet.py
pmelchior: But in this case, the class doesn't behave properly. Say, you have transform coefficients and create an instance with this static method: `starlet = Starlet.from_starlet(coeffs)`. What is `starlet.image` or `starlet.norm`?

pmelchior: Btw, here you can see again the problem I have with calling the transform coefficients `starlet`. It mixes up instances of the class `Starlet` with numpy arrays of the starlet coefficients.

scarlet/wavelet.py
```diff
@@ -0,0 +1,309 @@
...
+    @property
+    def norm(self):
+        """The norm of the seed wavelet in each wavelet level (not in coarse wavelet)"""
+        return self._norm
+
+
+    @property
+    def coefficients(self):
+        """Starlet coefficients"""
+        return fft._centered(self._starlet,
```
pmelchior, line 74: Is there a benefit of reshaping them here instead of returning `self._starlet`?

scarlet/wavelet.py
pmelchior: I suggest re-arranging a few items here:
* call this method `from_coefficients(coefficients, shape=None, starlet_shape=None)`
* don't keep reference to original image
* move the generation of an image from coefficients to `@property def image`.

The logic is the following. You normally instantiate the class with an image, but from then on the image itself is not needed any longer. By maintaining a reference you prevent the original from going out of scope, which is just waste of memory.

If you want to instantiate from a set of coefficients, use this method. In both cases, the image can/should be made only if needed, which would make this method faster as well.

scarlet/wavelet.py
herjy: I see the trouble. A Starlet object is a container for an image and its transform. It can either be instantiated from the image or from the starlet coefficients. If 'image` and 'starlet' are both provided, it is up to the user to make sure they know what they are doing. Alternatively, I could make it so that 'image' and 'starlet' (to be replaced by 'coefficients') are both optional parameters and mutually exclusive. I.e. you can declare a Starlet object from both ends, but for consistency, we avoid being able to provide both.

pmelchior: see my suggestion on refactoring below

scarlet/wavelet.py
herjy: Yes, if there has been some padding because of the way the coefficients were generated.

scarlet/wavelet.py
herjy: So what you want this to do is:
* Remove the 'starlet' argument from __init__
* compute the inverse in `from_coefficients`
* Declare a new Starlet object from the inverted image. 
Is that right? 
If so, no! Because the last bullet will call the starlet transform. But you don't need it because your coefficients already exist. This is supposed to be used when no image exist, only a set of coefficients!

scarlet/wavelet.py
pmelchior: Understood. This looks consistent with the `fft._centered` call in `from_starlet`, which I suggested to rename to `from_coefficients`. Please confirm because then you can avoid using the `_centered` method there and directly work in the coefficients, which would make this scheme transparent to the user:
```
starlet = Starlet(image)
coeffs = starlet.coefficients
starlet_ = Starlet.from_coefficients(coeffs)
```

scarlet/wavelet.py
pmelchior: No, not right.
* You can keep `starlet` (or better `coefficients`) in `init`, that just saves computing the transformation if you already have it.
* Simple set the coefficents and shapes, etc in `from_coefficients`. Don't create the inverse.
* Create the inverse only when requested in `@property image`

My main point is, the `Starlet` class does not need the image array, it's fundamentally defined by the coefficients.

scarlet/wavelet.py
herjy: This is how it works at the moment. I'll check if I can remove it.

scarlet/wavelet.py
pmelchior: For this to work, `init` needs to have optional `image` and `coefficient`, but one of them needs to be set. If you want to make them mutually exclusive or have coefficient takes preference is a matter of choice.

pmelchior: I agree with your last point. With my suggestion you can get rid of the staticmethod (which probably should be a classmethod anyway), and simply do `Starlet(coefficients=coeffs).image`

herjy: So optional image argument and then from_coefficients does not really need to exist anymore. __init__ will run the inverse transform?

pmelchior: Yes, but why would init run the inverse transform? You seem to want to cache the inverse instead of creating it on-the-fly. Why is that?

herjy: Because I don't see a reason to declare a Starlet object with a set of coefficients if it's not to get the inverse.

pmelchior: Yeah, I get the same impression. For one, your code, also needs to call the `image` property explicitly, like here: https://github.com/pmelchior/scarlet/pull/166/files#diff-0bef36ec53a572489b9a02f7bc87f501R37, so in terms of convenience your current approach is as good as mine. But yours needs to create the image before it can even create the class and it keeps a reference to `image`. The first is not necessary (because coefficients are all you need) and the second is inefficient memory-wise.  

My suggestion avoid both problem by creating `image` on-the-fly, which has the extra advantage of being automatically consistent with the coefficients. This way you e.g. could modify the coefficient without having the generate a new `Starlet` instance solely for the purpose of getting a new inverse.

scarlet/wavelet.py
```diff
@@ -28,91 +28,82 @@ def __init__(self, image, lvl = None, starlet = None, direct = False):
...
+        self._direct = direct
+        if coefficients is None:
+            if image is None:
+                raise InputError('At least an image or a set of coefficients should be provided')
             else:
-                self._starlet = self.transform()
+                if len(image.shape) == 2:
+                    self._image = image[np.newaxis, :, :]
+                else:
+                    self._image = image
```
pmelchior, line 41: You're still keeping a reference to the image. Why? This prevents the original from going out of scope.

scarlet/wavelet.py
```diff
@@ -28,91 +28,82 @@ def __init__(self, image, lvl = None, starlet = None, direct = False):
...
+                # Original shape of the image
+                self._image_shape = self._image.shape
+                # Padding shape for the starlet transform
+                if lvl is None:
+                    self._lvl = get_starlet_shape(image.shape)
+                else:
+                    self._lvl = lvl
+            self._coefficients = None
```
pmelchior, line 49: I think this can be simplified. I'd always compute the coefficients in `__init__` and use  `@property coefficients` only to protect them from being overwritten.

The idea is: regardless of how the class is instantiated (with an image or coefficients), at the end of `__init__` it has the coefficients and forgets the image (if it ever had it).

scarlet/wavelet.py
```diff
@@ -28,91 +28,82 @@ def __init__(self, image, lvl = None, starlet = None, direct = False):
...
     def image(self):
         """The real space image"""
-        return self._image
+        if self._image is None:
+            rec = []
+            for star in self._coefficients:
+                rec.append(iuwt(star))
+            return np.array(rec)
+        else:
+            return self._image
```
pmelchior, line 80: I'd remove the if/else in favor of always computing the inverse transform. Unless there is a good reason the store and maintain it.

scarlet/wavelet.py
```diff
@@ -28,91 +28,82 @@ def __init__(self, image, lvl = None, starlet = None, direct = False):
...
     def coefficients(self):
         """Starlet coefficients"""
-        return fft._centered(self._starlet,
-                             [self._image_shape[0], self._starlet_shape[-3], *self._image_shape[-2:]])
+        if self._coefficients is None:
+            if self._direct == True:
+                return self.direct_transform()
+            else:
+                return self.transform()
+        else:
```
pmelchior, line 95: I'd move the computation of the coefficients to init and simply return them here

scarlet/source.py
herjy: It returns a nan. I suspect a division by zero.

scarlet/wavelet.py
herjy: Then, I think what should happen is that the element (image or coefficient) the object is instantiated with should not be kept in memory: Once the inverse is computed, I don't really need the coefficients anymore..

scarlet/wavelet.py
herjy: So, as you've seen, my whole initial plan was to have a rather rigid structure for the class to prevent computing transforms and inverse when it's already been done. I surrender to the argument of memory. The coefficients and images are in this class only to be transformed and modify. There is therefore no reason to keep them as we do in the Fourier class where it is useful to keep the psf and a few other operators in memory. I'll make it always compute

scarlet/source.py
pmelchior: Yes, I'm also getting weird results from the thresholded coefficients. Something isn't right, I think, fixing the step sizes just masks a more profound problem.

pmelchior:
So, overall, the new setters make this behave much better, thanks for that. I've added a few stylistic corrections, but I'm now primarily worried about the behavior of the code. When I try this, it doesn't seem to fit well at all, e.g. with shifts in color that I don't understand.

scarlet/__init__.py
```diff
@@ -8,5 +8,6 @@
 from .frame import *
 from .observation import *
 from .blend import *
+from .wavelet import *
```
pmelchior, line 11: Please remove this line. We're not importing `fft` into the package namespace either

scarlet/wavelet.py
```diff
@@ -0,0 +1,295 @@
...
+
+    Attributes:
+        expression -- input expression in which the error occurred
+        message -- explanation of the error
+    """
+
+    def __init__(self, message):
+        self.message = message
+
+def mad_wavelet(image):
```
pmelchior, line 281: This should be method of `Starlet`, then it could directly access its own coefficients. From what I can see, this is only used in this context anyway, so a free method is not needed.

scarlet/source.py
```diff
@@ -395,6 +396,101 @@ def __init__(self, frame, sky_coord, observations):
...
+    and the morphologies are initialised as ExtendedSources
+    and transformed into starlet coefficients.
+    """
+
+    def __init__(
+        self,
+        frame,
+        sky_coord,
+        observations,
+        noise,
```
pmelchior, line 412: I don't think it is necessary to force the user to compute this. Simply compute the transform, compute the MAD of the coefficients and you're good. cf `ExtendedSource`, where we also make the noise threshold computations ourself.

scarlet/wavelet.py
herjy: I'm closing this as it is delt with below.

scarlet/__init__.py
herjy: done

scarlet/wavelet.py
herjy: I disagree. This is a rather robust way of estimating the noise (assuming gaussianity) in a noise-dominated image, so it could easily be used on any images regardless of whether this image is to be modeled by wavelets.

scarlet/source.py
herjy: I thought I had a good reason to do that but now it eludes me.

herjy: Oh, I know, that's in case the noise is not gaussian. Here you could imagine giving a full rms map.

scarlet/wavelet.py
herjy: Fixed

scarlet/wavelet.py
herjy: Fixed

scarlet/wavelet.py
herjy: Fixed by updating image/coeffs whenever each other's setter is called.

scarlet/wavelet.py
herjy: Outdated

herjy:
I think the issue with the quality of the fit is now due to the thresholding. I am not completely sure about what happens there yet.

scarlet/wavelet.py
pmelchior: That makes no sense to me. The method is in a module called `wavelet` and internally computes the starlet coefficients (not some coefficients, starlet coefficients). This is redundant and not at all general. It should be a starlet method or property.

scarlet/source.py
pmelchior: If the noise is not Gaussian, we're in deeper trouble (the log likelihood isn't right then). If it's spatially variable it would be stored in `observation.weights`, it is passed to the source constructor already. It's part of the job of `...Source` to set up the parameters properly, so I don't see a good reason to make the user perform very implementation-specific computations.

pmelchior:
I'm OK with testing / tuning starlet coefficient step size and thresholding schemes in a follow-up PR. But we can't really make use of the implementation without that.

scarlet/wavelet.py
herjy: Say I have an image for which I want to get the noise standard deviation. I would find it super awkward to have to do this: `sigma = Starlet(image).mad_wavelet` instead of `sigma = mad_wavelet(image)`. I'm fine with having this function in another module, but I don't see where else it could go.

pmelchior: I strongly prefer `sigma = Starlet(image).mad`. It also avoids computing the starlet transform twice: once to get `noise` to call `StarletSource`, and the another time to set the starlet transform for that `StarletSource`.

herjy: Nope, that wouldn't work like that, would it. Noise is computed from the transform of the image. The sources are set by the transform of the sources as initialised by `init_extended_source`, so it doesn't cut on anything. the Starlet(image).mad makes it look weird and obscure, when `sigma = mad_wavelet(image)` looks more straightforward and transparent: That's a function that gives you the mad.

pmelchior: OK, fair point

scarlet/source.py
```diff
@@ -395,6 +396,104 @@ def __init__(self, frame, sky_coord, observations):
...
+
+        sed = Parameter(
+            sed,
+            name="sed",
+            step=partial(relative_step, factor=1e-2),
+            constraint=PositivityConstraint(),
+        )
+
+        noise = mad_wavelet(observations[0].images) * \
+                np.sum(observations[0]._diff_kernels.image**2, axis = (-2,-1))**0.5
```
pmelchior, line 463: `observations[0]` should be `observations[obs_idx]` to be consistent with the initialization of `image_morph`

pmelchior: Also `np.sqrt` is a lot faster than `**0.5`

scarlet/source.py
```diff
@@ -395,6 +396,104 @@ def __init__(self, frame, sky_coord, observations):
...
+        # Threshold in units of noise
+        thresh = 5 * np.sqrt(np.sum((sed*noise) ** 2))
```
pmelchior, line 465: Is this really not configurable from the outside? The factor 5 looks a little dubious to me.

scarlet/source.py
```diff
@@ -395,6 +396,104 @@ def __init__(self, frame, sky_coord, observations):
...
+        morph = self.transform.coefficients*0
+        # wavelet-scale norm
+        starlet_norm = self.transform.norm
+        #One threshold per wavelet scale: thresh*norm
+        thresh_array = np.zeros(morph.shape) + thresh
+        thresh_array = thresh_array * np.array([starlet_norm])[..., np.newaxis, np.newaxis]
+        # We don't threshold the last scale
+        thresh_array[:,-1,:,:] = 0
+
+        morph_constraint = ConstraintChain(*[L0Constraint(thresh_array), PositivityConstraint()])
```
pmelchior, line 479: Hmm: are starlet coefficients strictly positive for positive sources?

scarlet/source.py
herjy: I originally had it as an optional parameter in the initialisation. This is the threshold in units of noise so here, I'm thresholding at 5*sigma. In practice I have hardly ever set it to a different value, so I can put it back as an optional parameter .

scarlet/source.py
herjy: Yes, this generation of wavelets was specially designed for this. But it doesn't mean that a sparsest representation cannot be achieved that includes negative coefficients. The actual way to do this is to do sub-iterations to find the point at the intersection of the set of positive-valued real images and the set of images with a starlet decomposition in the desired L1(0) ball.  
It's not super clean to do it this way.

scarlet/source.py
pmelchior: I'm OK if this is the most suitable setting, but I can envision us tinkering with it, too. Your call.

scarlet/source.py
pmelchior: If strictly non-negative coefficients result in strictly non-negative images, our approach also yields a feasible solution. Is this condition satisfied by starlets?

herjy: Yup! By design of this specific familly (second generation starlets)
Screen Shot 2020-05-13 at 4 50 55 PM
How cool would it be to to be able to type and execute small codes directly inside comment bubbles?!

scarlet/source.py
herjy: Optional parameter to `Starlet_source` then.

scarlet/source.py
pmelchior: Excellent!

PR #167, opened 2020-05-07
Display

pmelchior:
Minor consistency tweaks in display: 
* origin is always lower (stops weird image flips in docs)
* and models are always unmasked.

On the last point: Observations have masks, our models do no. Also, masking at zero creates a lot of visual noise exactly where nothing relevant happens. The exact location of the zero crossing is noise dominated and largely irrelevant (maybe somewhat meaningful for PSF models). Also, not every model is thresholded at zero, so the plotting code imposes a constraint, which the actual source model may not share.

### DIFF
--- a/docs/0-quickstart.ipynb
+++ b/docs/0-quickstart.ipynb
@@ -276,9 +276,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "scrolled": false
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "scarlet.display.show_sources(sources, \n",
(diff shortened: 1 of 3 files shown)


fred3m:
This will have a lot of merge conflicts with the direct gradients since it touches a lot of the same lines, and I already made most of these changes already in my branch. So would you mind if I just incorporate your changes into that branch (which should be ready for review today).

pmelchior:
Yes, but so far I have not seen `origin` being set in your branch

fred3m:
I did it a different way. I like your 'origin=lower' solution, whereas I just offset the `extent` of `imshow` by -1/2, which produces the same result. My next commit should have that change implemented.

pmelchior:
Covered by #165

PR #168, opened 2020-05-08
flat monotonicity weighting

pmelchior:
* Implements flat monotonicity weighting over all three neighbor pixels, which reduces radial streaking
* Allows the user the set the monotonicity weighting scheme for each source model.

Plus some spring cleaning ... I removed the special code for nearest-neighbor monotonicity, which we haven't used anymore for more than 1 year.

### DIFF
--- a/scarlet/operator.py
+++ b/scarlet/operator.py
@@ -9,26 +9,6 @@ from . import interpolation
 from .cache import Cache
 
 
-def _prox_strict_monotonic(X, step, ref_idx, dist_idx, thresh=0):
-    """Force an intensity profile to be monotonic based on nearest neighbor
-    """
-    from . import operators_pybind11
-
-    operators_pybind11.prox_monotonic(X.reshape(-1), ref_idx, dist_idx, thresh)
-    return X
-
-
-def _prox_weighted_monotonic(X, step, weights, didx, offsets, thresh=0):
-    """Force an intensity profile to be monotonic based on weighting neighbors
-    """
-    from . import operators_pybind11
-
-    operators_pybind11.prox_weighted_monotonic(
-        X.reshape(-1), weights, offsets, didx, thresh
-    )
-    return X
-
-
 def sort_by_radius(shape, center=None):
     """Sort indices distance from the center
 
@@ -70,16 +50,26 @@ def sort_by_radius(shape, center=None):
     return didx
 
 
-def prox_strict_monotonic(shape, use_nearest=False, thresh=0, center=None):
+def _prox_weighted_monotonic(X, step, weights, didx, offsets, min_gradient=0):
+    """Force an intensity profile to be monotonic based on weighting neighbors
+    """
+    from . import operators_pybind11
+
+    operators_pybind11.prox_weighted_monotonic(
+        X.reshape(-1), weights, offsets, didx, min_gradient
+    )
+    return X
+
+
+def prox_weighted_monotonic(shape, neighbor_weight="flat", min_gradient=0, center=None):
     """Build the prox_monotonic operator
 
     Parameters
     ----------
-    use_nearest: `bool`
-        Whether to use the nearest pixel to the center for comparison
-        (`use_nearest=True`) or use a weighted combination of all
-        neighbors closer to the central pixel (`use_nearest=False`).
-    thresh: `float`
+    neighbor_weight: ['flat', 'angle', 'nearest']
+        Which weighting scheme to average all neighbor pixels towards `center`
+        as reference for the monotonicty test.
+    min_gradient: `float`
         Forced gradient. A `thresh` of zero will allow a pixel to be the
         same value as its reference pixels, while a `thresh` of one
         will force the pixel to zero.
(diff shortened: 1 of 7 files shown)


scarlet/operator.py
```diff
@@ -597,17 +570,20 @@ def getRadialMonotonicWeights(shape, useNearest=True, minGradient=1, center=None
...
             indices = maxIndices * cosNorm.shape[1] + columnIndices
             indices = np.unravel_index(indices, cosNorm.shape)
-            cosNorm[indices] = minGradient
+            cosNorm[indices] = 1
             # Remove the reference for the peak pixel
             cosNorm[:, px + py * shape[1]] = 0
         else:
+            if neighbor_weight == "flat":
+                cosWeight[cosWeight != 0] = 1
+
```
fred3m, line 586: Overall I like this except that since "flat" is the default there is no reason to go through all of the trig to build the `cosWeight` version of `cosNorm`. TBH I'm not sure why we were doing this every time before. So I think it would be faster to only calculate the angles and `cosWeight` if `neighbor_weight=="angles"`. For flat weights you can just check that `dx**2 + dy**2 < x**2+y**2`, where `x` and `y` are the coordinates of each pixel.

scarlet/source.py
```diff
@@ -207,10 +213,10 @@ def init_extended_source(
             morph, 0, center=center, algorithm="sdss"
         )
 
-    if monotonic:
+    if monotonic is not None:
         # use finite thresh to remove flat bridges
-        prox_monotonic = operator.prox_strict_monotonic(
-            morph.shape, use_nearest=False, center=center, thresh=0.1
+        prox_monotonic = operator.prox_weighted_monotonic(
+            morph.shape, neighbor_weight=monotonic, center=center, min_gradient=0.2
```
fred3m, line 219: This doesn't cause the initial footprints to be larger? I think that was the reason that we used strict monotonicity before in the initialization.

fred3m:
Is this waiting on me to approve? I just want to double check because I see that you pushed changes but it doesn't seem to cache the entire prox (as you said you were doing on slack) and it still uses trig functions for the flat weights.

pmelchior:
No, it's stuck with me to look at the modification you suggested.

scarlet/source.py
pmelchior: That's a misnomer: even `operator.prox_strict_monotonic` with `use_nearst=False`, used weighted monotonicity: https://github.com/pmelchior/scarlet/blob/master/scarlet/operator.py#L115

scarlet/operator.py
```diff
@@ -534,122 +505,82 @@ def diagonalsToSparse(diagonals, shape, dtype=np.float64):
...
-    """Create an operator to constrain radial monotonicity
-    This version of the radial monotonicity operator selects all of the pixels closer to the peak
-    for each pixel and weights their flux based on their alignment with a vector from the pixel
-    to the peak. In order to quickly create this using sparse matrices, its construction is a bit opaque.
-    """
-    import scipy.sparse
+    # Find the difference between each pixels angle with the peak
+    # and the relative angles to its neighbors, and take the
+    # cos to find its neighbors weight
+    dAngles = (angles.flatten()[:, None] - relativeAngles.T).T
```
herjy, line 559: I thought from a peak you were only ever looking at the three neighobouring pixels downstream of the light profile. Why then compute weights relative to all the pixels in the image?

pmelchior: I'm also deciphering what's going on in this piece...

fred3m: I think that this is a remnant of the way we used to do constraints with ADMM. Looking back on it, as is usually the case looking back at old code, if I was to do it again I would do it differently but the whole thing was built in a few stages.

For Remy, we actually flattened the data into a vector and performed constraints and transforms using matrices so that the linear algebra interpretation was literal. So we needed to build a matrix to apply the constraints and this originally returned a large sparse matrix that we applied to the model with a real dot product. But that was sssslllloooowwww, so we started to create the matrix as a scipy sparse diagonal matrix by converting everything into an array that represented the diagonals and could be used to initialize the sparse matrix. That's the time where the majority of the code comes from, converting data and weights into diagonals for a large sparse matrix. This is probably the last piece of the codebase that hasn't been refactored yet since then (because it always worked).

Looking at the code now for the first time in years, none of this makes sense for what we're doing now, even if we use weights. So fixing it seems beyond the scope of this PR and I'm fine with you just skipping it for now and fixing it as part of a new issue (or assigning it to me to cleanup since it's my mess). But if you want to fix it then you should just probably stop looking at the old code altogether and rewrite it from scratch, since what we're doing now is so much simpler (especially your new flat weights method).

pmelchior: +1 I'm ready to throw the towel. I can't even get the nearest-neighbor weighting to work...

herjy: I got the linear algebra formulation of your operations, I'm just surprised by the size of the matrix, but if it's deprecated then let's make it a ticket and solve it in a later PR.

fred3m: Ah, actually, looking at the code now it isn't doing what I think that you think it's doing.

The `diagonalizeArray` function takes a 2D array, flattens it, and then makes copies of it in the locations of its 8 neighbors (since we don't know a prior which 3 neighbors are closer to the center). So all of the diagonalized arrays are 8xN, where N is the total size of the image. So for each of a pixels 8 neighbors it calculates a relative angle between the pixel and the neighbor, which is what that operation shows, and then a mask is applied to the 5 neighbors that are furthest from the peak.

So this is actually a fairly efficient way of building things up to the line that you highlighted if we restrict ourselves to numpy vectorized operations (I won't say the _most_ efficient way, but it might be close). But since we're using C++ to apply the operator we might as well build it there too, where we can be more efficient for flat weights. But for now if @pmelchior modifies https://github.com/pmelchior/scarlet/blob/flat-mono/scarlet/operator.py#L558-L582 to something like

```
if neighbor_weight == "flat":
    weight = np.zeros(relativeDist.shape, dtype=np.float64)
    weight[~invalidPix] = 1
    return weight
cosWeight = np.cos(dAngles)
cosWeight[invalidPix] = 0
cosWeight[mask] = 0
...
```

that's along the lines of what I was thinking about in my comment during review.

herjy: That's what I would expect it to do, but I don't see where it says that the size is reduced to the 8 nearest pixels. Also, it seems weird to me to compute the angle relative to those when we know the values these angles will take (or am I missing something?)

fred3m: In https://github.com/pmelchior/scarlet/blob/flat-mono/scarlet/operator.py#L541-L549,
`xArr` and `yArr` are "diagonalized" to be `8xN` pixel arrays. So `dx`, `dy` are also dimension `8xN` and so is `relativeAngles`. So in https://github.com/pmelchior/scarlet/blob/flat-mono/scarlet/operator.py#L557 `relativeAngles` is `8xN` and `angles` is projected to be the same shape.

Issue #172, opened 2020-05-16
compute coads only once for initialisation


herjy:
Sources are initialised by building coadds weighted by a pre-computed sed and applying a set of morphological constraints on the coadds to get a morphology. This means that the coad has to be computed for each sed, resulting in computing several time the same multiplication weigthed differently depending on background rmses. This is generaly cheap, but suboptimal. On top of this, the weighting ensures that the amplitude of the morphologies correspond to the amplitude of the seds. However, the morphologies are now normalised to their max at unity so knowing this we could just create one coadd weighted by the background levels and initialise all sources on that. The current initialisation is robust to a general implementation where sources are not max-normalised, but do we really need to support this? 

Also the current architecture requires that we pass raw observations to the function that build coads. In the multiresolution case, this would lead to either:
* performing an interpolation of all low-res images for each source (highly suboptimal!) or
* having to create synthetic `Observation` for each `LowResObservation` that contains the interpolated images at high-res. Possible, but I find it inelegant.

I propose to refactor the initialisation and have the coadd created only once. The morphologies would be initialised with a max normalisation on top of the rest of the current initialisation.

herjy:
I think the best way to do this is to make a rather big overhaul of the initialisation. I'm thinking of an initialisation class that would take coordinates, frames, observations and a list of source classes to perform the initialisation. This would be, I think, more user-friendly than the weird for loop with `if` statements to initialise each source one by one.

fred3m:
I'm not sure that I agree with your interpretation, but I might be missing something. When we build the detection coadd we weight all bands with the SED of each source not because we want the amplitude to be correct (since we normalize it away when we factor in the SED and PSF) but because sources have power in different parts of the spectrum. So a faint red source requires a different coadd than a bright blue source, where the faint source needs more weight given to the redder bands to properly initialize its morphology (and sometimes to even be detected at all).

With that being said, I've been looking at the detection coadds lately for sources of different colors and they aren't _that_ different. So it might be that for most sources using a common coadd will save enough time to be worth the slight degradation in initialization, whereas other times it will be inappropriate. So my recommendation would be to create a `detection_coadd` parameter that can be passed to a source. If `detection_coadd` is `None` then the coadd will be rebuilt for a given source, otherwise it will use the joint `detection_coadd`. Then it will be up to our custom initialization schemes (which will be different for different use cases) to decide which sources can use a joint coadd and which ones will need their own custom coadd.

fred3m:
[...]

Issue #171, opened 2020-05-15
multi-resolution initialisation


herjy:
So far the multiresolution initialisation is done outside scarlet, but I'm going to formalise it now. It's not a lot of code, but it can be done either of two ways: 
* I think it's better done by passing an optional parameter to tell the current initialisation classes to call an init function that will generate the multiresolution coadd and then do the detection as it would normally do (at the moment I generate the coadd outside scarlet and feed it with an artificial observation that contains the coadd: not ideal).
* The other option is to dedicate an initialisation class and a function to perform these tasks, but I see this repeat a lot of the code of existing classes and find it inelegant. 
Since @pmelchior normally has the last word when it comes to architecture, I'll let you decide.

herjy:
Another key point is how to do the coadd. Currently, I interpolate low-res images to high-res and sum everything together after rescaling the total flux in each image. This works in practice, but the interpolation introduces correlated noise that worries me a little. Since it is just initialisation, this does not need to be a super mathematically correct solution, but I'm worried about the the estimation of the background cutoff. 

Another option I have tested is to denoise all images before coadding them and adding gaussian noise afterwards so as to not mess pieces of the code that might try to estimate the noise in a noise-free image (although I am not sure any of these are in the intialisation path). I have a very robust wavelet filtering method implemented in my branch that does that really well. 

On top of that, this ignores the psf completely. I don't think it's paramount especially given that the hr image where the model lives is in the coadd, but it could be explored.

fred3m:
A while back I tried to put initialization code in scarlet and @pmelchior balked, believing that it isn't our responsibility to make too many decisions for end users. So I created the [scarlet_extensions|https://github.com/fred3m/scarlet_extensions] package to handle my init functions and any other methods that I may need in the future for scarlet processing.

So one thing that we could do, to include the work from @pmelchior and his students, @herjy , and myself, is to make a more formal extensions package with different initialization schemes and other code that is too specific to the use case for the main `scarlet` repo but will be useful for other researchers.

However, if the way that you initialize sources doesn't have any decision making and will be the same (up to a few configuration parameters) every time, for every user using multi-resolution data, then it probably is suitable for inclusion in the main scarlet repo.

herjy:
[...]

PR #173, opened 2020-05-19
Init multi

herjy:
This PR adds a few features for initialising sources as discussed in issues #171 and #172. 
* Initialisation can now be performed by a function that takes `Frame`, `Observation`s and matching lists of coordinates and source types. 
* Initialisation can be performed on a coadd image either weighted by the seds of each source (1 coadd per source) in each observation, or inverse variance weighted or flat-weighted for the case where wavelet filtering is used (1 coadd total).
* The previous API for initialising sources is still supported and features the possibility to provide a coadd and background cutoff directly.
 * `Starlet` has a new `filter` method that acts as a very power de-noising filter.


docs/tutorials/multiresolution.ipynb
```diff
@@ -257,14 +251,11 @@
...
-    "                           symmetric=False, \n",
-    "                           monotonic=True, \n",
-    "                           obs_idx=1)\n",
-    "    for i in range(ra.size)\n",
-    "]\n",
+    "# For better detections, turn obs in sources. New initialisation scheme.\n",
+    "sources = initialise(frame, \n",
+    "                     obs, \n",
+    "                     [(r, d) for r, d in zip(ra, dec)], \n",
+    "                     scarlet.Extended(thresh = 5, min_grad = 0.1), coadd_sed = False)\n",
```
pmelchior, line 258: I see this as a way to share the coadd, alright. But the API is clunky (see below the whole slew of empty code in `sources.py` that's needed to support this. It's also not transparent and impractical to the user. How would this look when there's a mix of source types? You'd have to manually sync the coordinate and source lists (plus all of their respective argument). In essence, you're trying to hide the very thing that needs to be exposed to the user, namely what they are fitting.

scarlet/initialisation.py
```diff
@@ -0,0 +1,136 @@
+from .source import *
+from .interpolation import interpolate_observation
+from .observation import Observation
+from .wavelet import Starlet
+
+def build_initialisation_coadd(observations, wavelet = False):
```
pmelchior, line 6: American English, please!

scarlet/initialisation.py
```diff
@@ -0,0 +1,136 @@
...
+    if wavelet is True:
+        coadd /= np.max(coadd)
+        bg_cutoff = 0
+        return coadd, bg_cutoff
+    coadd /= jacobian
+    # thresh is multiple above the rms of detect (weighted variance across channels)
+    bg_cutoff = np.sqrt((weights ** 2).sum()) / jacobian
+    return coadd, bg_cutoff
+
+def initialise(
```
pmelchior, line 67: That function should not exist, and I'll show where the concept breaks.

scarlet/initialisation.py
pmelchior: But I do think it's a good idea to bundle initialization methods here, to clean up `sources.py`

scarlet/initialisation.py
```diff
@@ -0,0 +1,136 @@
...
+        coadd, bg_cutoff= build_initialisation_coadd(observations, wavelet = wavelet)
+    else:
+        coadd, bg_cutoff = None, None
+    source_list = []
+    for i, coord in enumerate(sky_coords):
+        if len(sources) == 1:
+            source = sources[0]
+        else:
+            source = sources[i]
+        if type(source) in (Point, Random):
```
pmelchior, line 123: For this to work, you'd need to know *every* source type users ever wants to initialize. That is unmanageable if you want to also allow users to create their own custom source types. Which we do.

scarlet/source.py
```diff
@@ -59,17 +61,68 @@ def get_psf_sed(sky_coord, observation, frame):
...
-        sed /= observation.frame.psf.image.max(axis=(1, 2))
+        sed /= observation.frame.psf.image.sum(axis=(-2, -1))
 
     if frame.psf is not None:
-        sed = sed * frame.psf.image[0].max()
+        sed = sed * frame.psf.image[0].sum()
 
     return sed
 
+def build_sed_coadd(seds, bg_rmses, observations):
```
pmelchior, line 78: this functions call all go to `initialization.py` (note the spelling...) as well

scarlet/source.py
```diff
@@ -295,6 +331,15 @@ def init_multicomponent_source(
...
 
+class Random:
+    """ Class used to instantiate a RandomSource class from its kwargs.
+    """
+    def __init__(self, observations = None):
+        """Sets a *Source with all its arguments"""
+        self.kwargs = observations
+    def __call__(self, *args):
+        """Sets a *Source with all its arguments"""
+        return RandomSource(args[0], observation = self.kwargs)
```
pmelchior, line 342: That is bad code because it serves almost no purpose. It simply provides a mechanism for you to run the external initialization, which is delayed from the actual instantiation of the source class. And its unmanageable again unless you want to force the user to write that empty code as well.

The delayed initialization (e.g. with a common coadd) can be a good idea, but this way isn't.

scarlet/source.py
```diff
@@ -317,19 +362,31 @@ def __init__(self, model_frame, observation=None):
...
             morph, name="morph", step=relative_step, constraint=constraint
         )
 
         super().__init__(model_frame, model_frame.bbox, sed, morph)
 
+class Point:
+    """ Class used to instantiate a PointSource class from its kwargs.
+    """
+    def __call__(self, *args):
+        return PointSource(*args)
```
pmelchior, line 389: same as above, but even emptier

scarlet/initialisation.py
```diff
@@ -0,0 +1,136 @@
...
+        The minimum value in `detect` to include in detection.
+    """
+    try:
+        iter(observations)
+    except TypeError:
+        observations = [observations]
+    # The observation that lives in the same plane as the frame
+    loc = np.where([type(obs) is Observation for obs in observations])
+    # If more than one element is an `Observation`, then pick the first one as a reference (arbitrary)
+    obs_ref = observations[loc[0][0]]
```
pmelchior, line 28: That needs to be at least explained to the user. We used to have the `obs_idx` keyword exposed to the user, but this is completely implicit. I'm also not sure how this would work with a joint observation coadd

scarlet/source.py
```diff
@@ -451,21 +531,18 @@ def __init__(
             sky_coord,
             frame,
             observations,
-            obs_idx=obs_idx,
```
pmelchior, line 454: I'm not sure that this argument should be removed

pmelchior:
I agree with the problem you seek to address, namely that the initialization of the sources could be better if it was done delayed and simultaneous, e.g. for sharing a multi-resolution detection coadd that is expensive to produce.
One could also think of fitting for the SEDs of all sources once their morphologies are initialized because that's a linear solution.
Other ideas are probably conceivable as well.

Here's the crux: You need to know that the source wants to be (or rather: what it can fit; what are feasible solutions) to choose an initialization that works well. This is why the initialization procedure is part of the `XYZSource` class implementation. I realize that you haven't changed that aspect, but you've hidden it in the `initialise(...)` method, which has a pretty opaque and practically unmanageable implementation.

I much prefers @fred3m's proposal of optionally feeding a detection coadd to the initialization. It's a minimal change to the current master and solves your immediate problem.

docs/tutorials/multiresolution.ipynb
herjy: It is not that much more code in sources.py. We could get rid of the "Setter" classes if we decided to change the API and stopped supporting the current intialisations actually. Also, I'm open to suggestions that could replace these classes. I couldn't find a better way.

Look at the initialisation in quickstart to see what it looks like when there are multiple sources. Honestly between this and and the `for ... if ... elif ... else` of the current implementationI find this one easier to. read and understand. Indeed, I am not fond of the fact that coordinates and sources have to match, but you also have to do this in the current implementation and on top of that you have to place the right coordinates as argument of EACH source. I don't see how any of your comments do not apply to the current initialisation scheme.

scarlet/initialisation.py
herjy: Yes, that I am not happy with either, but I can make it differently to accommodate generic sources

scarlet/source.py
herjy: Yes, it's the bit that I find inelegant. I could get rid of it, and only rely on the source classes, but it would not be able to support the new API anymore.

scarlet/initialisation.py
herjy: This is tricky. Actually, I put that here only because the current code suggests that we might feed multiple `Observation`s but I hardly a case where one would need to fit more than one `Observation` and here I make the distinction between `Observation` and `LowResObservation`. I think we should expect lists that don't have more than one `Observation`.

herjy:
But it's already what we do! Not a lot is hidden and providing a coadd is actually possible with this update.

docs/tutorials/multiresolution.ipynb
herjy: The idea behind this is that I see in the future the possibility of passing a catalog that contains source type. At some point, we [or someone] will have to make an informed decision about what type of source each source. With this implementation, we could imagine a detection method that spits coordinates and source types to be fed directly to the initialise function.

herjy:
In the end, I'm ok to drop the `initialisation` function and construct the coadd to feed to the initialisations of the sources, but I have yet to see a convincing argument in favour of the current implementation. I really don't like the fact that for each source we have to provide the same set of arguments. External coadds will make it even worse.

scarlet/initialisation.py
fred3m: +1

scarlet/initialisation.py
fred3m: I can think of a few counter examples, for example multi-epoch observations, where there is a separate `Observation` for each epoch. So I prefer to require to set the observation index if the length of `observations` is greater than 1, if for no other reason than to make the user think about what they are doing.

scarlet/initialisation.py
```diff
@@ -0,0 +1,136 @@
...
+        if obs is obs_ref:
+            if wavelet is True:
+                images = Starlet(obs.images).filter()
+            else:
+                images = obs.images
+        else:
+            #interpolate low-res to reference resolution
+            images = interpolate_observation(obs, obs_ref.frame, wave_filter = wavelet)
+        n_images += len(images)
+        if wavelet is True:
```
fred3m, line 51: I'm not sure I like this as the parameter name to decide how to build the coadd. I would prefer to either make separate functions for different coadd types or use a more descriptive name for the different coadd types and pass that as either a "magic string" or (as I prefer) and enum.

scarlet/initialisation.py
fred3m: I agree. For reasons stated in #171 I think this goes against the current design pattern of scarlet. But something similar (and more general) would be useful to include in a separate repo that we make public to give users access to more specialized scripts.

scarlet/interpolation.py
```diff
@@ -563,6 +565,23 @@ def sinc_interp_inplace(image, h_image, h_target, angle, pad_shape = None):
     return sinc_interp(image, coord_hr, coord_lr, angle=angle)
 
 
+def interpolate_observation(observation, frame, wave_filter = False):
+    """ Interpolates the images in an observation on the grid described by a frame
```
fred3m, line 569: please add parameters and returns to the docstring

scarlet/source.py
```diff
@@ -59,17 +61,68 @@ def get_psf_sed(sky_coord, observation, frame):
     sed = get_pixel_sed(sky_coord, observation)
 
     # approx. correct PSF width variations from SED by normalizing heights
+    if observation._diff_kernels is not None:
+        sed /= observation._diff_kernels.image.sum(axis=(-2,-1))
+
+        return sed
```
fred3m, line 67: I don't think that using `sum` in this function as opposed to `max` will give better colors, in fact I'm almost certain that it won't. In general the PSFs are normalized to sum to 1 in each band (in fact it may even be a requirement), so in most cases the sed is not affected at all by the `sum` operation.

Recall that our SED and morphology exist in the model frame before convolution with the difference kernel. This means that when we build the model in the first iteration we are convolving with the difference kernel, which will change the amplitude of the SED. That is the effect that the `max` normalization is meant to (mostly) cancel, and has significantly improved our initial colors. It also helps our initial amplitude estimates, but it isn't perfect because a pixel is also affect by the PSF from its neighboring pixels. So we could improve the amplitudes by weighting some subset of the central pixels by the PSF, which I think we did at some point and realized that it did more damage to the initial colors. With this algorithm the colors are right because it adjusts the central pixel for variations in the width of the PSF in each band.

But I do agree that if the difference kernels are already built then we could just use the _maximum_ value of the difference kernel to adjust the SED.

herjy: Indeed, the colours are worse. 
Intuitively I don't like the max. because the central pixel is not just affected by the max pixel of the psf but all the pixels of the psf multipled by the image's pixels around the central pixel. But I surrender to the fact that in practice it works well.

scarlet/initialisation.py
herjy: I might actually drop this all together. It doesn't bring anything. The results are pretty much the same.

scarlet/initialisation.py
herjy: Won't these have to be `LowResObservation`s? If it requires interpolation they will.

scarlet/source.py
herjy: I have delayed pushing my PR because I'm testing the use of the diff_kernel here. I don't understand why, but I get better colours  when I do this :
`sed *= observation._diff_kernels.image.sum(axis=(-2,-1))`
And in the case of multi-resolution, the difference is huge! Using this leads to much better results. 
For single observations, I've only tested onn the quickstart and at e_rel = 1.e-7, it takes 193 iterations with this scheme against 237 iterations with the current implementation in master.

docs/tutorials/multiresolution.ipynb
herjy: Removed

scarlet/initialisation.py
herjy: Kept and americaniSed.

scarlet/initialisation.py
herjy: removed

scarlet/source.py
herjy: Removed

scarlet/initialisation.py
herjy: Solved with an optional obs_idx. None leads to selecting the first one, because if you care which frame you are fitting into, you know what you are doing and otherwise, users probably won't care or hav only one `Observation`.

scarlet/initialisation.py
herjy: Name chaged to `filter_coadd`

scarlet/interpolation.py
herjy: Done

scarlet/source.py
herjy: Solved. The sed is estimated on a low resolution image for a model at high resolution, so the integration is wrong and needs a factor h**2. Now this is replaced by this: 
```
sed *= observation._diff_kernels.image.sum(axis=(-2,-1))
return sed/observation.h ** 2
```
The max of the diff kernel doesn't work because diff kernels are not uni-modal, mootonically decreasing from the center out and might have several pixels at the max.

fred3m: Please use a parameter to choose the normalization method. If the user chooses "max" then use the same normalization as before (since that wasn't acting on the difference kernel, it should still be valid). Until I can verify that this improves initialization on the HSC test dataset (or at least doesn't make it worse) I need to be able to use an initialization that I know works.

docs/tutorials/multiresolution.ipynb
```diff
@@ -230,14 +231,7 @@
     "# This implementation is a bit of a hack and will be refined in the future\n",
```
fred3m, line 231: Is this still true after this PR?

scarlet/initialization.py
```diff
@@ -0,0 +1,438 @@
...
+def init_extended_source(
+    sky_coord,
+    frame,
+    observations,
+    coadd=None,
+    bg_cutoff=None,
+    thresh=1,
+    symmetric=True,
+    monotonic=True,
+    min_grad=0.2,
```
fred3m, line 63: Please make the default `min_grad=0.0`, since we know that higher values do not work well for many objects with the new `flat` monotonicity weighting.

Also, please change `monotonic` to allow the weight type to be specified, the same as `ExtendedSource`. This is make sure that our initialization and constraints are consistent.

scarlet/source.py
```diff
@@ -506,25 +231,28 @@ def __init__(
...
         sky_coord,
         observations,
-        obs_idx=0,
+        coadd=None,
+        bg_cutoff=None,
         thresh=1.0,
         monotonic="flat",
         symmetric=False,
         shifting=False,
+        min_grad=0.2,
```
fred3m, line 240: Please change `min_grad` to `0.0`.

scarlet/source.py
```diff
@@ -617,12 +346,14 @@ def __init__(
...
         observations,
-        obs_idx=0,
+        coadd=None,
+        bg_cutoff=None,
         thresh=1.0,
         flux_percentiles=None,
         symmetric=False,
         monotonic="flat",
         shifting=False,
+        min_grad = 0.2,
```
fred3m, line 356: Please change `min_grad` to `0.0`.

scarlet/initialization.py
```diff
@@ -0,0 +1,438 @@
...
+    else:
+        if bg_cutoff is None:
+            raise AttributeError(
+                "background cutoff missing! Please set argument bg_cutoff"
+            )
+    # Apply the necessary constraints
+    center = frame.get_pixel(sky_coord)
+    if symmetric:
+        morph = operator.prox_uncentered_symmetry(
+            coadd*1, 0, center=center, algorithm="sdss" # *1 is to artificially pass a variable that is not coadd
```
fred3m, line 102: why `coadd*1`? Why is the multiplication needed?

scarlet/source.py
herjy: Good point! Setting min_grad to 0 by default brought back the loss at 31652.619806746377

scarlet/initialization.py
herjy: Replaced by .copy()

scarlet/initialization.py
herjy: Done

fred3m:
Actually, one last comment. Could you please change the default `min_gradient` to `0.1`? Otherwise the stamps in blends with a lot of sources are much too big, since they can grow arbitrarily large without decreasing in flux at all. I tested using `min_gradient=0.1` on highly elliptical galaxies and galaxies with a large faint disk and (at least for ground based images) this works much better. I'd change it and push an update myself but then I'd cause merge conflicts on your end.

herjy:
I get where this is coming from but the notebooks don't do as good with 0.1. For instance here is the notebook with min_grad at 0.1 (e_rel = 1.e-7 to let it go longer).
Screen Shot 2020-05-25 at 4 43 16 PM


> Actually, one last comment. Could you please change the default `min_gradient` to `0.1`? Otherwise the stamps in blends with a lot of sources are much too big, since they can grow arbitrarily large without decreasing in flux at all. I tested using `min_gradient=0.1` on galaxies with high eccentricity and galaxies with a large faint disk and (at least for ground based images) this works much better. I'd change it and push an update myself but then I'd cause merge conflicts on your end.

fred3m:
Hmm, I can't tell by looking at it that it's any worse than the current master from the [docs](https://pmelchior.github.io/scarlet/_images/tutorials_point_source_11_0.png).



Are the residuals larger, and if so, is it only because they're running longer? I'm just surprised because all of the HSC images that I tested show improvement with little to no effect on the residuals.

fred3m:
Wait, that's the point source tutorial, so none of those are initialized as extended sources. So this shouldn't affect them at all.

herjy:
I have something very Diffrerent:
Screen Shot 2020-05-25 at 5 21 38 PM

Also, yes, two sources are Extended ones in this notebook

pmelchior:
The noise level in this image is unknown and also not set from `sep`'s background. The quality of this model is questionable in all cases.

The setting of `min_gradient=0.1` re-establishes what we have done in the past. It's true that this setting is a compromise between compactness and completeness. Different cases will respond differently. Please set `min_gradient=0.1`, let Jarvis run the tests, and merge.

fred3m:
Sigh, looks like the notebook logL is failing. I'm fine with removing this line now anyway because the new regression tests will supersede it and should be ready by next week.

herjy:
Grrrr sorry, I need to find why my pytest won't run the notebook correctly

PR #183, opened 2020-06-05
Randomly skip a parameter update to prevent degeneracy

fred3m:


fred3m:
There doesn't seem to be any difference, either positive or negative. But it can't hurt.

herjy:
Do you still see the oscillations during optimisation?

PR #186, opened 2020-07-03
Implement scarlet_test for regression testing

fred3m:


pmelchior:
There are no issues about this PR on our end, but the workflow for `scarlet_test` should be streamlined. Right now, someone needs to manually create an PR, and Fred needs to manually run a git fixup.

I suggest that we check how to deploy the test results (a few numbers and several png figures) to a web server, e.g. an amazon bucket, and pull the figures from there by a dynamic website. This way, no PR in `scarlet_test` is needed, and we can load whatever branches we want to compare from the website.

fred3m:
I looked into it and neither Google nor AWS have a free tier that includes cloud data any more (after the 1st 12 months). However, AMW has a DynamoDB that I think will work for our purposes and is actually the way I was originally thinking about this. So as long as I can get it setup properly all of the measurement data and residuals will be pushed to tables in the database that we can query in our docs. I'll look into this next week.

.travis.yml
```diff
@@ -3,6 +3,15 @@ python:
   - 3.6
 notifications:
   email: false
+env:
+  global:
+    - USER="fred3m"
+    - EMAIL="fred.moolekamp@gmail.com"
+    - GH_REPO="github.com/fred3m/scarlet_extensions"
```
pmelchior, line 10: I'd prefer if we called this variable differently. it might confuse our future selves because this *isn't* the scarlet repo

.travis.yml
```diff
@@ -28,6 +37,22 @@ install:
   - python setup.py install
 # Run test
 script:
+  - echo ${TRAVIS_PULL_REQUEST}
+  - echo ${TRAVIS_BRANCH}
+  - pwd
+  - cd ..
+  # Clone scarlet_extensions (we do it here because we might need to checkout a branch)
+  - git clone https://${GH_REPO}
```
pmelchior, line 45: if you change `GH_REPO` above, it would have to be updated here as well

docs/index.rst
```diff
@@ -19,6 +19,7 @@ Getting Started
    1-concepts
    tutorials
    api/scarlet
+   regression
```
pmelchior, line 22: nice that this is included publicly. I'm not sure anyone would use it besides us, but it speaks to our desire for transparency.

scarlet/display.py
```diff
@@ -225,7 +225,7 @@ def show_scene(
     panels = sum((show_model, show_observed, show_rendered, show_residual))
     if linear:
         if figsize is None:
-            figsize = (3 * panels, 3 * len(list(sources)))
+            figsize = (5 * panels, 5)
```
pmelchior, line 228: I'm not sure the larger figures will work as well in the docs, but let's find out

PR #193, opened 2020-07-15
closes #185

### LINKED ISSUE #185: Programmatic versioning
It would be nice to have internal versioning so that `scarlet.__version__` does what you would expect it to do. It would be double nice to have that versioning track the GitHub tagged version (or commit).

I've been pretty happy with [versioneer](https://github.com/warner/python-versioneer) for solving these issues. I could probably be convinced to help getting it setup if the developers would be interested.

pmelchior:
consistent versioning scheme

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ install:
   - conda info -a
 
   # Replace dep1 dep2 ... with your dependencies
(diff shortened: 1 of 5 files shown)


pmelchior:
Yay! Travis fails with this error: `ModuleNotFoundError: No module named 'autograd'`. I don't understand why, it's still on pypi.

pmelchior:
Closed in favor of #197

PR #197, opened 2020-07-22
Use scm for versioning

fred3m:
This isn't intended to be merged. It is just an update to the `version` branch with a more robust versioning scheme.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,10 +4,12 @@ build/
 var/
 dist/
 *.egg-info
+*.eggs
 *.so
 .ipynb_checkpoints/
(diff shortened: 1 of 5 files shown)


pmelchior:
Is this meant as replacement of #193 or in addition to it? If mean as a replacement, then we should consider your comment that `display` should be imported as module in `__init.py__`.

fred3m:
Sure, I wasn't sure if that is what we had settled on. I think that I made all of the changes that you made, plus importing display now.

pmelchior:
Fixes #185

PR #201, opened 2020-08-05
added Observation parameter hooks; closes #200

### LINKED ISSUE #200: Observation with parameters
Current only sources have parameters. But sometimes we have to adjust the mapping between the observational frames and the model frame, e.g. when the astrometry is off.

The way to do that is to allow `Observation` to declare parameters, and to pass them from `Blend` as `Observation.get_loss(model, *parameters)`.

Specifically, an astrometric offset could be corrected by using a Fourier-space shift (same thing we do in `ImageMorphology`) when we do the convolution. Another possible but more delicate parameter would be the PSF model itself.

This PR is only to create the framework to create and pass observation parameters.

pmelchior:
Minimal addition to `factored` to allow for observational parameters

### DIFF
--- a/scarlet/blend.py
+++ b/scarlet/blend.py
@@ -94,7 +94,9 @@ class Blend(CombinedComponent):
         alg_kwargs: dict
             Keywords for the `proxmin.adaprox` optimizer
         """
-        X = self.parameters
+        X = self.parameters + tuple(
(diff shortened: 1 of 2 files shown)


pmelchior:
This PR is just for the hooks. PSF parameters need a redesigned PSF model first.

PR #202, opened 2020-08-05
astropy WCS

pmelchior:
This PR make our WCS treatment (more) consistent. It should go in after #191 because it's forked of from `factored`.


scarlet/frame.py
```diff
@@ -91,40 +93,45 @@ def psf(self):
...
-            if self.wcs.naxis == 3:
-                coord = self.wcs.wcs_world2pix(*sky_coord, 0, 0)[:2]
-            elif self.wcs.naxis == 2:
-                coord = self.wcs.wcs_world2pix(*sky_coord, 0)
-            else:
-                raise ValueError(
-                    "Invalid number of wcs dimensions: {0}".format(self.wcs.naxis)
-                )
-            return tuple(int(c.item()) for c in coord)
+            pixel = np.array(self.wcs.world_to_pixel_values(sky)).reshape(-1, 2)
```
herjy, line 105: Actually we should consolidate this with the functions in `resampling.py`, which are supposed to deal with this.

scarlet/frame.py
```diff
@@ -91,40 +93,45 @@ def psf(self):
...
+            # y/x instead of x/y:
+            pixel = np.flip(pixel, axis=-1)
+        else:
+            pixel = sky
 
-        return tuple(int(coord) for coord in sky_coord)
+        if pixel.size == 2:  # only one coordinate pair
```
herjy, line 111: Why would there be more than one pair? Also, if there is one pair, won't the size be 2?

pmelchior: This method now allows passing (N,2) arrays so the N transformations are done in one go.

fred3m:
Can you change the base of this to merge into `factored` the same way that you did for #201? That made it much easier to review, and then you can just merge all of the changes together later today.

scarlet/frame.py
```diff
@@ -34,10 +36,10 @@ def __init__(self, shape_or_box, channels, wcs=None, psfs=None, dtype=np.float32
         self.channels = channels
 
         if wcs is not None:
-            import astropy
-
             assert isinstance(wcs, astropy.wcs.WCS)
-        self.wcs = wcs
+            self.wcs = wcs.celestial  # only use celestial portion
```
fred3m, line 40: Nice, I didn't know that this existed!

scarlet/interpolation.py
```diff
@@ -546,26 +551,43 @@ def sinc_interp_inplace(image, h_image, h_target, angle, pad_shape = None):
...
+    coord_hr = (
+        np.array(
+            [
+                np.array(range(ny_hr.astype(int))) - (ny_hr - 1) / 2,
+                np.array(range(nx_hr.astype(int))) - (nx_hr - 1) / 2,
+            ]
+        )
+        / h_image
+        * h_target
+    )
```
fred3m, line 586: gross `black`, just gross. <sigh>

Issue #195, opened 2020-07-17
Starlet example notebook


pmelchior:
The current notebook I don't find that impressive. So, I agree with you that it should have its own notebook with a dedicated example, but we should attempt to keep the complications around detection to a minimum in that example notebook. So, I will keep the notebook, but we should find a better example.

_Originally posted by @pmelchior in https://github.com/pmelchior/scarlet/pull/191/review_comment/create_

herjy:
I think I have one. I'll push it after this PR is done.

PR #203, opened 2020-08-13
Starlet init

herjy:
A small PR limited to the starlet models that formalises the possibility to initialise Starlets with an SED and without bounding box (other than the image boundaries).
The wavelet tutorial notebook was also updated with a more compelling example.
This addresses issue #195.


scarlet/spectrum.py
```diff
@@ -45,12 +45,12 @@ class TabulatedSpectrum(Spectrum):
         1D bounding box for focation of the spectrum in `frame`
     """
 
-    def __init__(self, frame, spectrum, bbox=None):
+    def __init__(self, frame, spectrum, bbox=None, step=1e-2):
```
pmelchior, line 48: I'm not a fan of this option. The logic is the following: either you provide an array, which will be converted to a `Parameter` *with default everything*, or you provide your own `Parameter` instance. The middle ground is murky. Someone might want to change the constraint, but leave the step size etc. It's really hard to allow for some changes as options without changing others.

We could say, well, then we make all `Parameter` options (constraint, prior, step, even current gradient etc.) accessible to the user, but then you write interfaces that look a lot like a `Parameter` instantiation.

scarlet/initialization.py
```diff
@@ -169,29 +169,19 @@ def trim_morphology(center_index, morph, bg_thresh):
...
-    sky_coord,
+def morph_init(sky_coord,
     frame,
     observations,
+    seds,
     coadd=None,
     coadd_rms=None,
-    thresh=1,
-    symmetric=True,
+    symmetric=False,
```
pmelchior, line 178: Don't simply change defaults for the initialization. `ExtendedSource` works well with it. If you don't like that, specify `symmetric=False` when you call this function.

scarlet/initialization.py
```diff
@@ -169,29 +169,19 @@ def trim_morphology(center_index, morph, bg_thresh):
     bbox = Box.from_bounds((bottom, top), (left, right))
     return morph, bbox
 
-
-def init_extended_source(
-    sky_coord,
+def morph_init(sky_coord,
     frame,
     observations,
+    seds,
```
pmelchior, line 175: This should be an optional parameter because it only is necessary when `coadd is None`. If so, you can check with an `assert` that `seds is not None`

scarlet/initialization.py
```diff
@@ -237,6 +228,42 @@ def init_extended_source(
...
+):
+    """Initialize the source that is symmetric and monotonic
+    See `ExtendedSource` for a description of the parameters
+    """
+    if not hasattr(observations, "__iter__"):
+        observations = (observations,)
+
+    # determine initial SED from peak position
+    # SED in the frame for source detection
+    spectrums = [get_psf_sed(sky_coord, obs, frame) for obs in observations]
```
pmelchior, line 253: `spectra`, please!

scarlet/initialization.py
```diff
@@ -237,6 +228,42 @@ def init_extended_source(
...
+    spectrum = np.concatenate(spectrums).reshape(-1)
+
+    morph, bg_rms, center_index = morph_init(sky_coord,
+                                             frame,
+                                             observations,
+                                             spectrum,
```
pmelchior, line 259: I'm quite sure that this needs to be `spectra` because you need the amplitudes in the same order as `observation` and not concatenated. That's the whole reason for calling `get_psf_sed` separately for all observations above.

scarlet/source.py
```diff
@@ -119,24 +121,28 @@ def __init__(
             Multiple of the backround RMS used as a
             flux cutoff for morphology initialization.
         """
+        if spectrum is None:
+            spectrum = get_pixel_sed(sky_coord, observations)
```
pmelchior, line 125: This is the same problem again, I think. These are concatenated spectra if more than one observation is present

scarlet/source.py
```diff
@@ -119,24 +121,28 @@ def __init__(
...
+            coadd=coadd,
+            coadd_rms=bg_cutoff,
+            symmetric=False,
             monotonic="angle",
             min_grad=min_grad,
             starlet_thresh=starlet_thresh,
         )
-        spectrum = TabulatedSpectrum(model_frame, sed, bbox=bbox[0])
+
+        spectrum = TabulatedSpectrum(model_frame, spectrum, step = 1.e-6)
```
pmelchior, line 141: This is where you should create a special `Parameter` instance with the small step size.

pmelchior:
No big issues. I'm looking forward to the updated starlet notebook.

One thing, when we're done with the review, please "squash and merge" to avoid the images from the notebook in 7b82b63 ending up in the master branch.

pmelchior:
Where do we stand on this PR?

scarlet/source.py
herjy: Addressed

scarlet/initialization.py
herjy: Not relevant anymore.

scarlet/initialization.py
herjy: Yup, that was a mistake.I mixed up sources.

scarlet/spectrum.py
herjy: I changed it by declaring the spectrum a Parameter object in the init_starlet_source

scarlet/source.py
herjy: Not relevant anymore.

pmelchior (approved):

Issue #177, opened 2020-05-27
Variable box sizes and/or centers for scarlet


pmelchior:
Currently box sizes and positions are determined at initialization. We know that this can be optimized by adjusting box sizes and centers on the fly with the current best estimate of the source morphologies. We used to have this capability and needed to switch it off with the new Adaprox optimizer.

Since then we have gain most of the capability back:
* The groundwork is laid by #158. The reason for getting rid of the box adjustments is that the optimizer stores multiple versions of the gradients internally (called `m`, `v`, and `vhat`). The warm-start PR allow us to control them as they are now attached to `Parameter`. 
* Our improved code for `Box` from #165 allows us to easily extend all the image-like arrays.
* #168 fixed a massive memory issue. In the process, `MonotonicityConstraint` now makes a cache lookup based on the shape of the `morph` argument and computes the proximal operation for any shape it hasn't cached yet.

Missing still are 

1. method to reshape the `_data`, `m`, `v`, `vhat` members of the `morph` parameter in `FactorizedComponent`
2. a decision criterion when to do so
3. control logic to run the decision and reshaping methods

Items 1 and 3 are straightforward. As we keep finding, 2. is not. We used to compute the average flux along the edges of the box and compare it to the noise level in the image. This get complicated with multiple observations. We don't store the effective noise level `bg_cutoff` from the initialization, and we normally only compute it for one `Observation`, which may not be the one that would benefit the most from growing the box.

Suggestions welcome!

pmelchior:
Oh, I forgot: The capability above are identical to those needed to shift the box if the center pixel is not optimal. This is most immediately needed for source with `MonotonictyConstraint`. The only difference is the decision criterion.

PR #204, opened 2020-08-13
PSF as model

pmelchior:
This PR brings 3 related updates.

* it unifies `PSF` to allow for images and functional representations. this treatment was so far only marginally consistent. In addition, a PSF can now expose its parameters, which allows them to be considered as observation-level parameters (via #200).
* it corrects inaccuracies of the spectrum initialization using a single pixel. this update should provides a better initial logL in most test cases.
* it provides a `compact` option for `ExtendedSource` that initializes the source morphology with the pixels of the model psf in a minimal box. this option is more stable (and appropriate) for minimally resolved galaxies.


scarlet/psf.py
```diff
@@ -79,190 +107,88 @@ def f(X):
...
 
-    This is the same as `~scarlet.Observation` except that the matching
-    algorithm uses a precalculated deconvolution kernel and the
-    images are deconolved using that kernel. A separate class exists
-    to make it less likely that the user executes this operation
-    more than once, which will have adverse affects on the
-    `images` property.
-    """
+    def get_model(self, *parameters, offset=None):
+        assert offset is None, "ImagePSF cannot be offset"
```
herjy, line 191: But it can with FFT! Why not implement it?

scarlet/source.py
```diff
@@ -44,7 +46,7 @@ def __init__(self, model_frame, observations=None):
         if observations is None:
             spectrum = np.random.rand(C)
         else:
-            spectrum = get_best_fit_seds(image[None], observations)[0]
+            spectrum = get_best_fit_spectra(image[None], observations)[0]
```
herjy, line 49: Ha! I wanted to change these in my PR, but realising that it crashed the tests, I renounced.

herjy:
Also, I thought we said we would allow the diff kernel to  shift instead of the psf, which is not what this PR does unless I missed it. It generally can not be done analytically though.

pmelchior:
This PR offers the option for us in the future to implement PSF fitting, should the need arise.

The shifting operation you refer to is to correct a mismatch in the observation WCS, which we should treat as an unknown affine transformation applied to the entire observation frame.

herjy:
I'm not sure we should. Why not just use the PSF shift to do that. Otherwise, if you for some reason a user lets the psf centre as a free parameter and the affine transformation of the grid free, you end up solving for 4 parameters that are degenerate.

pmelchior:
Yes, we cannot prevent the user from attempting something stupid.

herjy:
Well, in this case, yes, you can, by not adding a redundant functionality.

pmelchior:
These are different purposes. You may still want to be able to fit for an unknown PSF model even if you don't care about the WCS, or be so sure of the PSF model that you trust it to update the WCS. Both are valid, just not at the same time.

scarlet/morphology.py
```diff
@@ -154,8 +151,9 @@ def __init__(self, frame, center):
 
     def get_model(self, *parameters):
         center = self.get_parameter(0, *parameters)
-        spec_bbox = Box((0,))
-        return self.psf(*center, bbox=spec_bbox @ self.bbox)[0]
+        box_center = np.mean(self.bbox.bounds[1:], axis=1)
+        offset = center - box_center
+        return self.psf.get_model(offset=offset)  # no "internal" PSF parameters here
```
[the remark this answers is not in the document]
pmelchior: I agree in principle, but it's tied up with #177. I'd like to have the ability to change the boxes for all sources, not just for point sources, so it's better to treat them the same.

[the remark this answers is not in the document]
herjy: That's fair. I just raised it because I found that it was extra bad for point sources. When shifting point sources to fit quasars, we found that the psf quickly ends up being chipped by the edge of the box. Feel free to resolve.

scarlet/psf.py
```diff
@@ -1,72 +1,100 @@
...
+        """Normalize to PSF image in every band to unity
+        """
+        sums = image.sum(axis=(1, 2))
+        if isinstance(image, Parameter):
+            image._data /= sums[:, None, None]
+        else:
+            image /= sums[:, None, None]
+        return image
+
+    def prepare_param(self, X, name):
```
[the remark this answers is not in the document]
pmelchior: Agreed. Will be done in next commit

scarlet/psf.py
pmelchior: Because of the artifacts that arise, it isn't a good idea to shift an `ImagePSF`. That said I will implement it even if we generally should discourage model PSFs from images.

scarlet/psf.py
```diff
@@ -1,72 +1,100 @@
...
 from .parameter import Parameter
 from .fft import Fourier
 from .observation import Observation
 
 
-def moffat(y, x, alpha=4.7, beta=1.5, bbox=None):
-    """Symmetric 2D Moffat function
+class PSF(Model):
+    @abstractmethod
+    def get_model(self, *parameter, offset=None):
```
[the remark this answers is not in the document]
pmelchior: For the PSF, `offset` is a condition of the realization, not a parameter of the model. For example, a `GaussianPSF` has a width as `Parameter`, but the center of the PSF is by construction in the middle of the box. The `ImagePSF` from LSST have the same construction. 

On the other hand, a `PointSource` has a center parameter, but it's not shared by other PointSources, and certainly not by the underlying `PSF`.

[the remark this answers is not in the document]
fred3m: It's been a while, but I thought that part of the idea was to have the ability to allow the offset of the PSF to shift slightly in order to make up for misalignments in the PSF between bands. Did I misunderstand something or am I just making that up?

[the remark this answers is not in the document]
pmelchior: No, you are right about that, but I'm thinking of a different mechanism for that. A PSF mismatch would be an `Observation`-level parameter, i.e. `Observation` gets another parameter `offset`, which it exposes to `Blend` through its `parameter` property. So, the PSF stays as it is (because offset isn't a PSF parameter unlike e.g. `sigma`), instead the observational frame gets offset. I'm not sure if we need this to be band-dependent, but I would rather define the PSF model to be centered, and treat the offset as part of the WCS.

This is different from actually allowing the PSF model itself to change, for instance by adjusting `sigma` of `GaussianPSF`. That would be a PSF parameter, exposed through `Observation.parameter`.

In both cases, we still need to ability to recompute the matching kernels. Next PR...

scarlet/psf.py
```diff
@@ -79,190 +107,88 @@ def f(X):
...
-
-    def normalize(self):
-        """Normalize to PSF image in every band to unity
-        """
-        sums = self._image.sum(axis=(1, 2))
-        self._image /= sums[:, None, None]
-        return self
+        shape = (len(alpha), boxsize, boxsize)
+        origin = (0, -(boxsize // 2), -(boxsize // 2))
+        self.bbox = Box(shape, origin=origin)
```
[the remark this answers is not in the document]
pmelchior: Not quite. The channel length is computed differently, so I will need to think about that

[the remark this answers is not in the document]
pmelchior: But I'm going to create `FunctionPSF`.

scarlet/initialization.py
```diff
@@ -48,106 +48,138 @@ def get_best_fit_seds(morphs, images):
...
+        index = np.round(pixel).astype(np.int)
+
+        psf = obs.frame.psf.get_model()
+        bbox = obs.frame.psf.bbox + (0, *index)
+        img = bbox.extract_from(obs.images)
+
+        # img now 0 outside of observation, psf is not:
+        # restrict both to observed pixels to avoid truncation effects
+        mask = img[0] > 0
+        psf = psf[:, mask]  # flattens array in last two axes
```
herjy, line 147: Why flattening?

scarlet/initialization.py
```diff
@@ -48,106 +48,138 @@ def get_best_fit_seds(morphs, images):
...
+        img = img[:, mask]
+
+        spectrum = (img * psf).sum(axis=1) / (psf * psf).sum(axis=1)
```
herjy, line 150: I'm not sure where that is from. This formula would be correct if two of the psfs here were replaced by the inverse Fourier transform of the inverse of the Fourier transform of the psf, which does not exist since our psfs are band-limited. What gives you the right to use the psf instead?

herjy: Couldn't we use the inverse of the psf in Fourier with reconditioning?

scarlet/initialization.py
```diff
@@ -157,27 +189,71 @@ def trim_morphology(center_index, morph, bg_thresh):
...
     right = center_index[1] + boxsize // 2
     bbox = Box.from_bounds((bottom, top), (left, right))
     morph = bbox.extract_from(morph)
-    bbox = Box.from_bounds((bottom, top), (left, right))
     return morph, bbox
 
 
+def init_compact_source(
+    sky_coord, frame, observations,
+):
```
herjy, line 208: Maybe a docstring would help. I suspect users who declare new sources will want to use or at least look at initialisation functions.

scarlet/initialization.py
```diff
@@ -238,20 +321,48 @@ def init_extended_source(
...
+                    psf.shape[0] - shape[0],
+                    psf.shape[1] // 2 - shape[1] // 2,
+                    psf.shape[2] // 2 - shape[2] // 2,
+                ),
+            )
+            psf = bbox_.extract_from(psf)
+
+            # if model PSF is constant across bands (as it should) then factor
+            # is constant as well
+            factor = (morph[None, :, :] * psf).sum(axis=(1, 2)) / (psf * psf).sum(
```
herjy, line 353: Redundant code with `get_psf_spectrum`

scarlet/psf.py
```diff
@@ -1,268 +1,243 @@
...
+                    self._f(self._Y - offset[0], s)[:, None]
+                    * self._f(self._X - offset[1], s)[None, :]
+                    for s in sigma
+                ),
+                axis=0,
+            )
+        # use image integration instead of analytic for consistency with other PSFs
+        return normalize(psfs)
+
+    def _f(self, X, sigma):
```
herjy, line 137: Not sure it needs to be a class method.

herjy (approved):
No major comment.

scarlet/initialization.py
```diff
@@ -48,106 +48,138 @@ def get_best_fit_seds(morphs, images):
...
+    return spectrum
 
-        seds.append(sed)
 
-    sed = np.concatenate(seds).reshape(-1)
-    return sed
+def get_minimal_boxsize(size, min_size=15, increment=8):
+    boxsize = min_size
+    while boxsize < size:
+        boxsize += 8  # keep box sizes quite small
```
fred3m, line 171: should be `boxsize += increment`

scarlet/initialization.py
fred3m: +1

scarlet/initialization.py
```diff
@@ -157,27 +189,71 @@ def trim_morphology(center_index, morph, bg_thresh):
...
-    bbox = Box.from_bounds((bottom, top), (left, right))
     return morph, bbox
 
 
+def init_compact_source(
+    sky_coord, frame, observations,
+):
+
+    # get PSF-corrected center pixel spectrum
+    spectrum = get_pixel_spectrum(sky_coord, observations, correct_psf=True)
```
fred3m, line 211: why not use `get_psf_spectrum` here? Isn't this exactly the case that function is meant to handle?

scarlet/initialization.py
```diff
@@ -157,27 +189,71 @@ def trim_morphology(center_index, morph, bg_thresh):
...
+
+    # adjust box size to conform with extended sources
+    size = max(morph_.shape)
+    boxsize = get_minimal_boxsize(size)
+    morph = np.zeros((boxsize, boxsize))
+    origin = (
+        center_index[0] - (morph.shape[0] // 2),
+        center_index[1] - (morph.shape[1] // 2),
+    )
+    bbox = Box(morph.shape, origin=origin)
```
fred3m, line 233: This whole block seems like a lot of extra work, unless I'm missing something. Shouldn't `morph` just be the PSF image in a box shifted so that it is centered on the source location?

scarlet/psf.py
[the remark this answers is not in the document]
fred3m: So `offset` is here in case there is a PSF that we know is not centered in the box? Do we have a use case for this?

fred3m (changes requested):
Looks good overall, I can't wait to see these changes in master! I left a few comments (nothing major) in addition to the open comments by Remy. So lets wrap those up and this should be good to go.

scarlet/psf.py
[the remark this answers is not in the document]
herjy: I just used to produce a makeshift image of a star.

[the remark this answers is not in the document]
herjy: And in general you will want to use it without having to specify sigma and instead use the current model.

scarlet/initialization.py
```diff
@@ -48,106 +48,138 @@ def get_best_fit_seds(morphs, images):
     return seds
 
 
-def get_pixel_sed(sky_coord, observations):
-    """Get the SED at `sky_coord` in `observation`
+def get_pixel_spectrum(sky_coord, observations, correct_psf=False):
```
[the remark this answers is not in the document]
pmelchior: I disagree. we have `get_psf_spectrum` for that very purpose. Her we get the spectrum of a single pixel. Hence the name.

scarlet/initialization.py
pmelchior: the formula is the amplitude of `img` when projected onto `psf`. In other words, it's the factor you have to multiply `psf` WITH to get `img` (if `img` had the same shape as `psf`).

pmelchior: leaving a comment in code now

scarlet/initialization.py
pmelchior: just what happens when you apply the mask (which is necessary)

scarlet/initialization.py
pmelchior: not at all. this isolated equation is formally the same, although it differs in being applied to 3D cubes instead of 2D.  However, many more operations need to be done here.

Just for explanation, `spectrum` assumes the source to have point-source morphology, otherwise `get_pixel_spectrum` has no defined operation. This `factor` corrects that by finding out how much (in terms of a scalar number) `morph` looks like the (model) `psf`.

Leaving code comments here, too

scarlet/psf.py
pmelchior: sure. `self.integrate`. Also, where else would you put it?

scarlet/component.py
```diff
@@ -163,7 +162,12 @@ def get_model(self, *parameters, frame=None):
             (Channels, Height, Width) image of the model
         """
         spectrum, morphology = self.get_models_of_children(*parameters)
-        model = spectrum[:, None, None] * morphology[None, :, :]
+        if len(morphology.shape) == 2:
+            model = spectrum[:, None, None] * morphology[None, :, :]
+        elif len(morphology.shape) == 3:
+            model = spectrum[:, None, None] * morphology
```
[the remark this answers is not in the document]
pmelchior: Think of 3D PSF cubes, where every band is normalized. If you look at `PointSource`, it derives from `FactorizedComponent`, but this PR expresses all PSFs as hyperspectral models. I didn't want to introduce a `NormalizedCubeComponent`...

scarlet/initialization.py
pmelchior: No, these two are different. The `get_psf_spectrum` measures the spectrum with a (target) PSF-shaped window function. `get_pixel_spectrum` only uses a single pixel, but then has to correct for PSF shape variations.

scarlet/initialization.py
pmelchior: Fair q. Our model psfs are *super* small (<~10 pixels), which is smaller than we allow any `ExtendedSource` to be. Hence the box size adjustment.

pmelchior:
Don't understand why I can't reply inline...

https://github.com/pmelchior/scarlet/pull/204#discussion_r502024113

Yes, it's called `PointSource`.

scarlet/initialization.py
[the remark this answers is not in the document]
herjy: Oh I understood it as spectrum in pixel units (image frame) by opposition to deconvolved units. But snice there is a psf correction it did not work, but I got the idea now.

scarlet/initialization.py
herjy: Which is ok for compact source, but a big assumption for extended ones.

scarlet/component.py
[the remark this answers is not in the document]
fred3m: Ok, fair enough.

scarlet/initialization.py
fred3m: I understand that, but I thought that for a point source `get_psf_spectrum` will give the exact spectral model. Is your argument that since a `CompactSource` is not necessarily a point source that using `get_psf_spectrum` would be introducing a larger error? Or am I still not understanding something?

fred3m (approved):

fred3m:
I approved the changes, I'd just like a clarification about what I'm missing in the comment above. https://github.com/pmelchior/scarlet/pull/204#discussion_r503376901

scarlet/initialization.py
pmelchior: It's still different. We _never_ use `get_psf_spectrum`. Even `PointSource` uses `get_pixel_spectrum` because it is more reliable in blends. If the aperture is (observation)PSF-sized, there can be a whole lot of light from neighboring coming in.

The point here is that for _isolated_ point sources, these two functions (with PSF correction) yield the same. In the isolated case `get_psf_spectrum` is superior because it uses a larger aperture is is thus more robust to noise.

PR #206, opened 2020-08-28
Streamlining

pmelchior:
This PR streamlines common tasks. 
* `Blend.fit` writes logger message with number of iterations and final logL
* `display.show_likelihood` shows our convergence plots from a fit `Blend` instance.
* `display.show_observation` shows the image and PSF in common norm and size (like `show_scene` but without sources)

In the process, I renamed `loss` to `log_likelihood` in both `Blend` and `Observation`, and fixed the stupid -1 for good.


scarlet/blend.py
```diff
@@ -164,6 +167,12 @@ def expand_grads(*X, func=None):
             **alg_kwargs
         )
 
+        logger.warning(
```
fred3m, line 170: This seems more like `info` than a `warning`.

fred3m (approved):
This looks good. The only change that I requested is to make logger output an `info` instead of a `warning` to show the final logL, since this isn't really indicative of something that went wrong and will pollute peoples consoles/terminals.

fred3m:
Of course, the docs are failing to build, as well as some changes that are needed in `scarlet_extensions` since you changed the name of `Blend.loss`. So either deprecate `Blend.loss` or fix `scarlet_extensions` using a branch with the same name.

pmelchior:
I will re-introduce `Blend.loss` to be backward compatible.

As for `logger.warning`, I am aware that it's not formally a warning, but it's the default visible value, and I think it information most people would like to see after `Blend.fit`.

PR #208, opened 2020-09-03
Invert the load order of the merged branches

fred3m:
One of the datasets needed to make the regression testing docs work is the list of merged branches, in order. This is stored in a DynamoDB table with the name of the branch and the index order acting as the unique keys for the table. The problem is by default git returns the branches in order with the newest first, which causes problems when the table updates on a new merge to master.

This PR fixes the problem by inverting the order of the merged branches on write (in scarlet_extensions) and requires a simple change here to invert the order again so that we correctly use the "last N commits" in the correct order.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ script:
   - git clone https://${TESTING_REPO}
   - cd scarlet_extensions
   - # Checkout the scarlet_extensions branch, if one is available (otherwise it will use master)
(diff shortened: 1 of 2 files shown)


fred3m:
Build passes: https://travis-ci.org/github/pmelchior/scarlet/builds/723922001

No rush on this, but this ticket should get merged to master before any other branches, otherwise they will mess up the regression docs again.

pmelchior (approved):

PR #209, opened 2020-09-15
Versioneering

fred3m:
Strip out setuptools_scm for a more stable native implementation to use SCM tags for version control.


pmelchior (approved):
I'm not a fan of having to accommodate the LSST stack on our end, but if it doesn't break out stuff, so be it.

fred3m:
I get that, but this is really making up for a shortcoming of `setuptools_scm`. The problem is they don't allow for non-version tags at all, so even if I wanted to just implement something on the lsst side (which was my original plan) I can't because when `setuptools_scm` tries to parse the non-version tags it can't and throws an error.

But this solution is better anyway because it's only a few lines of code to implement what takes them an entire obfuscated package that does God-only-knows what.

PR #213, opened 2020-10-12
Fix bug in branch names

fred3m:
This PR fixes an issue that was created when porting the tests
from scarlet_extensions to scarlet. Namely, I didn't want to have
to introduce the `git` package as a dependency so I attempted to
replicate the code on my own. That did not go so well...
So this PR re-implements the git solution that was working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,9 @@ install:
 
 # Run test
 script:
-  - pytest tests
+  #- pytest tests --branch ${TRAVIS_BRANCH}
+  # Temporary to test changes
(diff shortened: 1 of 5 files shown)


pmelchior:
build failure

fred3m:
Ok, this is fixed now. It turns out the problem is that for some reason when travis runs on a branch it detaches it from the repo, so it is not actually running on any branch (which is why they give you the `TRAVIS_BRANCH` variable). So I modified the regression test to take that as a parameter and all seems to work again (although I can't be sure that the master branch will work until we actually merge this to master).

pmelchior (approved):

PR #226, opened 2020-12-17
Propagate downgrade and fallback parameters

fred3m:
This PR fixes the bug seen in DC2 data where a small edge source triggers a re-initialization but the `downgrade` and `fallback` parameters are not set, resulting in small, faint sources being deblended as point sources.

### DIFF
--- a/scarlet/initialization.py
+++ b/scarlet/initialization.py
@@ -980,6 +980,8 @@ def initSource(
                 maxComponents,
                 edgeDistance,
                 shifting=True,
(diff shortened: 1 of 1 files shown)


pmelchior:
I have no problems with the fix, other than that it still uses PointSource as last resort. My current branch will fix that.

pmelchior (approved):

fred3m:
Thanks for the quick review. Yes, this ticket is just a stopgap for the DM pipeline until your branch is merged.

PR #230, opened 2021-01-26
Bug fixes

fred3m:
This PR fixes a few bugs introduced in the last 2 PRs for untested features.

### DIFF
--- a/scarlet/component.py
+++ b/scarlet/component.py
@@ -90,10 +90,10 @@ class Component(Model):
             Image of the model to project.
             This must be the same shape as `self.bbox`.
             If `model` is `None` then `self.get_model()` is used.
-        bbox: `~scarlet.bbox.Box` or `~scarlet.frame.Frame`
-            The `Box` or `Frame` to project the model into.
-            If `frame` is `None` then the model is projected
-            into `self.model_frame`.
+        bbox: `~scarlet.bbox.Box`
+            The to project the model into.
+            If `bbox` is `None` then the model is projected
+            into `self.model_frame.bbox`.
 
         Returns
         -------
@@ -103,21 +103,15 @@ class Component(Model):
         # Use the current model by default
         if model is None:
             model = self.get_model()
-        # If the box has a dtype then use it, otherwise use
-        # the dtype of the model
-        if hasattr(bbox, "dtype"):
-            dtype = bbox.dtype
-        else:
-            dtype = model.dtype
         # Use the full model frame by default
-        if bbox is None or bbox == self.frame:
+        if bbox is None or bbox == self.frame.bbox:
             bbox = self.frame.bbox
             frame_slices = self._model_frame_slices
             model_slices = self._model_slices
         else:
             frame_slices, model_slices = overlapped_slices(bbox, self.bbox)
 
-        result = np.zeros(bbox.shape, dtype=dtype)
+        result = np.zeros(bbox.shape, dtype=model.dtype)
         result[frame_slices] = model[model_slices]
         return result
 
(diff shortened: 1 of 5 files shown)


scarlet/component.py
```diff
@@ -90,8 +90,8 @@ def model_to_frame(self, frame=None, model=None):
             Image of the model to project.
             This must be the same shape as `self.bbox`.
             If `model` is `None` then `self.get_model()` is used.
-        frame: `~scarlet.frame.Frame`
-            The frame to project the model into.
+        bbox: `~scarlet.bbox.Box` or `~scarlet.frame.Frame`
```
pmelchior, line 93: I have one qualm with that (which isn't really changed since we introduced this method): I don't think this method can every accept an actual `Frame` since there it cannot account for resolution/subpixel-position changes. As an option it is sort of disguised now, but code still allows it. I'd feel happier if it really only accepted `Box` here. Is there a case where another `Frame` is needed?

scarlet/initialization.py
```diff
@@ -508,7 +508,8 @@ def init_source(
         # deblending of the edge source, which for bright sources
         # may ruin an entire blend. So we reinitialize edge sources
         # to allow for shifting and return the result.
-        if source_shifting is False and hasEdgeFlux(source, edge_distance):
+        source.is_edge = hasEdgeFlux(source, edge_distance)
```
pmelchior, line 511: This piece of code shouldn't actually be in scarlet at all. It doesn't work as intended if I understand it correctly, and it can/should be set upstream.

pmelchior (commented):
minor clarification requested

herjy:
I have beef with line 277 of renderer.py. It should be this: 
```       
 center_y = np.floor(
            self._fft_shape[0] / 2.0 - (self._fft_shape[0] - model_frame.Ny) / 2.0
        ) + ((self._fft_shape[0] % 2) != 0) * ((model_frame.Ny % 2) == 0)
        center_x = np.floor(
            self._fft_shape[1] / 2.0 - (self._fft_shape[1] - model_frame.Nx) / 2.0
        ) - ((self._fft_shape[1] % 2) != 0) * ((model_frame.Nx % 2) == 0)
```
Should I put it in a separate PR?

pmelchior:
@herjy the current form https://github.com/pmelchior/scarlet/blob/master/scarlet/renderer.py#L277-L282 is from your code, and the `np.int` does an implicit `floor` plus a conversion to integer. I think it could be more explicit if we did that same. Functionally I can't see a difference.

herjy:
I agree, although my code works without the conversion. However, without the float it doesn't work. One of these `autograd.numpy as np` again

pmelchior:
Not with my version of autograd (1.3 if memory serves me). `autograd.numpy.int` is a valid function.

scarlet/component.py
fred3m: No, that's fine. I just wasn't sure how much this was used by either you or @herjy in your own scripts, and I didn't want to mess with the API if I didn't have to. But if you two are fine with it I'm happy to use a box exclusively.

scarlet/initialization.py
fred3m: That is true. I think that we can just remove this whole section now, since I should be able to just add the shifting parameter to the source class after it is created now. I don't think that was always true.

fred3m:
@herjy Yes, please put that in another PR so that you can test it.

herjy:
My autograd is up to date too. I don't understand. I updated my mac at the same time as I switch to python 3.9 and now everything is screwy! It's probably on me, nvm then.

pmelchior (approved):

PR #234, opened 2021-02-10
init with size of at least the PSF

pmelchior:
stabilize the morphology initialization at the faint/small side by making sure its at least as wide as the PSF

### DIFF
--- a/scarlet/source.py
+++ b/scarlet/source.py
@@ -136,7 +136,7 @@ class PointSource(FactorizedComponent):
 
 class CompactExtendedSource(FactorizedComponent):
     def __init__(
-        self, model_frame, sky_coord, observations, shifting=False,
+        self, model_frame, sky_coord, observations, shifting=False, boxsize=None
     ):
         """Compact extended source model
 
@@ -156,13 +156,15 @@ class CompactExtendedSource(FactorizedComponent):
             Observation(s) to initialize this source.
         shifting: `bool`
             Whether or not a subpixel shift is added as optimization parameter
+        boxsize: int or None
+            Spatial size of the source box
         """
         if not hasattr(observations, "__iter__"):
             observations = (observations,)
 
         # initialize morphology from model_frame psf
         assert model_frame.psf is not None
-        morph, bbox = self.init_morph(model_frame, sky_coord)
+        morph, bbox = self.init_morph(model_frame, sky_coord, boxsize=boxsize)
         center = model_frame.get_pixel(sky_coord)
         morphology = ExtendedSourceMorphology(
             model_frame,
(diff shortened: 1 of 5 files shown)


scarlet/source.py
```diff
@@ -189,9 +190,23 @@ def __init__(
         # retain center as attribute
         self.center = morphology.center
 
-    def init_morph(self, sky_coord, frame):
+    @staticmethod
+    def init_morph(frame, sky_coord, boxsize=None):
```
fred3m, line 194: I get why you made this a static message, but is there a reason that you changed the order of the arguments?

fred3m (approved):
I like it. It also satisfies Bob's (and likely other peoples) request that the box size can be set or fixed if desired (whether or not that is a good idea is another discussion).

scarlet/source.py
pmelchior: same order as the constructor

pmelchior:
I'd like to hold off with the merge until we tested Remy's random init idea because it can help in similar cases. I will play with it a bit.

pmelchior:
Actually, re `boxsize`. It's not exposed at the level of the constructor. If this is desired functionality, I could add it.

fred3m:
Yes, that would be preferred. I implemented a fix in LSST that added a config option and just switched off `_resize` for all of the morphologies, but the ability to turn it off (or specify a box size) would be great.

fred3m (changes requested):
Everything looks good except that you don't expose `resize` to the the sources, only the morphologies. So I would add a `resize` argument to `ExtendedSource` and the other source classes and pass that through to morphology, which already has a `resize` argument. Then this should be all set.

pmelchior:
I had a slightly different philosophy when writing this. The things that need to be exposed are those that cannot be altered otherwise: initialization foremost. However, anyone can switch on/off both resizing and shifting now _after the source has been created_.

fred3m:
Yes, but it's not trivial when given a list of sources and you don't know what type of source each one is. You have to iterate over all of the sources, check the source type, and descend down the components until you get to the morphology, and set resize to false. That procedure is also prone to break, perhaps unknowingly, when we change the source/morphology classes internally.

It's much simpler and less prone to hidden failure to be able to pass `resize` and `shift` to the `init_all_sources` method and have the models created in the way that you expect, where any breakage due to a change in API will be obvious.

fred3m (approved):

fred3m:
Looks good. It's interesting to see some of the changes in the residuals for the test dataset. Most of them look better, but some have strange color artifacts in the center. Do you know what's causing those?

pmelchior:
I just checked the residuals of the validation data set. The overwhelming majority looks better now, I've only found one case where this branch looks worse than the main branch (and that was almost certainly a star). I think we are OK to proceed.

PR #240, opened 2021-04-06
Mr paper

herjy:
just a few minor bug corrrections I've met while running sims and investigating our various issues.


pmelchior:
You need to take out the `leaky_prox` commit because that branch is experimental.

Issue #245, opened 2021-06-23
Deblend failure when weights=0 for one or more images?


entaylor:
I'm running into an issue when trying to deblend a set of images, where the data are incomplete in one or more images.

What i have done is to set the values of both images and weights to be zero where no good data exist.  
![missing_data_example](https://user-images.githubusercontent.com/12062965/123023815-6b0aaf80-d41b-11eb-8b3f-6d8f68702d46.png)

I am uploading an image to illustrate a problem case.  From left to right, the columns show: the imaging data; the weights associated with the imaging; the rendered models after deblending, obtained as observation.render(blend.get_model()).

Two things happen.  First, the fluxes associated with those images where data are missing blow up to very large numbers.  This is sort of understandable, in that the fluxes are unconstrained on the interval of positive numbers.  The second is the bigger problem: the fluxes associated with images that do have data are driven to zero: ~1e-20.  This shouldn't happen, because the fluxes here **can** be well constrained.

It seems superficially that the masking of bad/missing data via setting weights to zero isn't fully respected in the case where some large fraction of one or more images is bad/missing, with the bad behaviour somehow tied to an overall normalisation of the spectrum.  

I've also attached the same case, but excluding the first two images (where data are missing), to show the 'good' solution.  I could include in my pipeline a check for missing data (though i'm not sure what the right condition might be precisely: the centre pixel? Some fraction of the bounding box?), but i feel like the code should be able to cope with this case?
![missing_data_example_noproblem](https://user-images.githubusercontent.com/12062965/123025840-9c38af00-d41e-11eb-9a53-d44915a2fe72.png)

fred3m:
Thanks for passing this along, this isn't something that we've come across before, and we do run this on images with zero weighted regions. But I did a test run just now using three different configurations: the first configuration was the full image with the default weights, the second configuration had zero weights and zero image values for the bluest band from [:60, 40:] (the lower right corner), and the last configuration used zero weights and image values in the reddest band from [:60, 40:].

Original
<img width="929" alt="Screen Shot 2021-06-23 at 12 13 11 PM" src="https://user-images.githubusercontent.com/2251900/123132754-123c1680-d41d-11eb-8333-b4b73396aac0.png">

Blue zeros
<img width="931" alt="Screen Shot 2021-06-23 at 12 16 45 PM" src="https://user-images.githubusercontent.com/2251900/123132818-241db980-d41d-11eb-8491-27e3e3f9dabb.png">

Red zeros
[...]

PR #247, opened 2021-07-02
Spectrum initialization incorrect for strongly masked data

pmelchior:
As reported by #245,  non-sensical spectra arise from fitting data with strong masking, i.e. when sources are initialized in regions where one or more bands are largely unavailable. The counter-intuitive behavior is that the amplitude is very high in the masked bands, and essentially zero in the unmasked bands.

Two aspects contribute to this problem:
1. The initialization with the `set_spectra_to_match` method solves a linear problem that becomes ill-defined it a morphology vector has no (or almost no) overlap with the region of non-zero weights. If convolutions are in the forward path, this statement applies to the convolved morphology. In both cases, this yields a numerically unstable matrix inverse whenever the bulk of the source is in the masked region of a given band. The solver finds the spectrum amplitude for all components in each band, which means that a failure of one component in a band ruins the amplitude estimate for most/all components in that band (but other bands are unaffected). This is likely the cause for the counter-intuitive behavior. aa4ccd2 fixes this problem by only inverting the matrix for the sub-block of components that have sufficient non-zero product of morphology*weight.

2. The gradients used during the fit are quite volatile, which can lead to unexpected trajectories in parameter space. Here's an example of a `ExtendedSource` fit to data of a white source where the lower half of the reddest of 5 bands has weights=0: 

https://user-images.githubusercontent.com/1463403/124305843-a13ee200-db33-11eb-9b37-9ca9067b6169.mp4

After the initialization (which underestimates the amplitude in the reddest band and thus comes out in cyan), there's a strong red overcompensation. That's a really bad fit, so the model gets suppressed globally. When it returns it avoids the 5-band region in the top of the image. Because it has lower flux there, it cranks up the red amplitude to get the right intensity in that band (the bottom region will not object because it doesn't care about the red amplitude). In this noise-free data the problem eventually gets corrected but it takes a large number of iterations because there's no strong incentive in terms of the log-likelihood of the fit. This problem can only really be avoided by introducing more information to the model in terms of parameter constraints/priors. For instance, fitting the same data with a `PointSource` (or probably any other parameterized model) has no trouble getting the correct flat (=white) spectrum. Priors in the spectrum shape would likely help as well. 

The fits in #245 appear to use point-source models, so the fix for problem 1 above should be all that is needed. I write down problem 2 because it's pretty counter-intuitive, but not a bug in the code itself. It's a problem for the optimizer to find the correct values when any change in the non-parametric morphology yields a different mix of masked pixels in various bands, and any change of the spectrum will lead to changes in the morphology...

For completeness, the other two commit fix corner cases that I have triggered during testing the stuff above.

### DIFF
--- a/scarlet/renderer.py
+++ b/scarlet/renderer.py
@@ -12,7 +12,12 @@ class Renderer(Model):
     def __init__(self, data_frame, model_frame, *parameters):
         self.data_frame = data_frame
         self.model_frame = model_frame
+        # mapping of model to data frame channels
         self.channel_map = self.get_channel_map(data_frame, model_frame)
+        # properly treats truncation in both boxes
+        # needs to overwritten if frames have different resolutions
+        self.slices = overlapped_slices(data_frame.bbox, model_frame.bbox)
+
         super().__init__(*parameters)
 
     # renderer is a parameterized transformation function
@@ -81,11 +86,14 @@ class Renderer(Model):
 
 
 class NullRenderer(Renderer):
-    def __init__(self):
-        super().__init__(None, None)
+    def __init__(self, data_frame, model_frame):
+        super().__init__(data_frame, model_frame)
 
-    def __call__(self, model, *parameters):
-        return model
+    def get_model(*parameters):
+        def nothing(model):
+            return model
+
+        return nothing
 
 
 @primitive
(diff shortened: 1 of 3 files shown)


scarlet/renderer.py
```diff
@@ -81,11 +86,14 @@ def map_channels(self, model):
 
 
 class NullRenderer(Renderer):
-    def __init__(self):
-        super().__init__(None, None)
+    def __init__(self, data_frame, model_frame):
```
fred3m, line 89: Thanks for pushing this fix to make `NullRenderer` usable. I had a similar fix on my branch but hadn't pushed it yet.

fred3m (approved):
Looks good to me, this makes sense and should work. While it's outside the scope of this ticket, if you solve for the spectra in each band independently it should be less memory intensive.

pmelchior:
The solver runs independently in each band.

fred3m:
Yup, you're right. Not sure what I was thinking. 🤷

PR #248, opened 2021-07-19
Psf shift optim

herjy:
This PR introduces tha capability for PSF parameters to be part of the model by way of introducing a new renderer.


scarlet/blend.py
```diff
@@ -262,8 +262,9 @@ def _loss_func(self, *parameters):
         for observation in self.observations:
             n_obs_params = len(observation.parameters)
             obs_params = parameters[n_params : n_params + n_obs_params]
+            shift = parameters[-1]
             total_loss = total_loss - observation.get_log_likelihood(
-                model, *obs_params, noise_factor=self._noise_factor
+                model, *obs_params, shift, noise_factor=self._noise_factor
```
pmelchior, line 267: how is any of that necessary? just pass `obs_params`.

scarlet/renderer.py
```diff
@@ -237,6 +240,69 @@ def convolve(self, model, convolution_type=None):
...
+            print('zizizi',shift)
+            model_ = self.convolve(model_, shift=shift)
+            # adjust spatial shapes
+            model_ = match_shape(model_, self.data_frame, self.slices)
+            return model_
+
+
+        return transform
+
+
```
pmelchior, line 305: Seeing this implementation, I think I should change my stance on this being a separate class. It's much simpler to add a single `shift=None` parameter to `ConvolutionRenderer` instead of dubplication almost all of it's contents, for the one thing that really happens (namely line 262)

scarlet/renderer.py
```diff
@@ -237,6 +240,69 @@ def convolve(self, model, convolution_type=None):
...
+    def get_model(self, *parameters):
+
+        def transform(model, *parameters):
+            # restrict to observed channels
+            model_ = self.map_channels(model)
+            # convolve observed channels
+            shift = self.get_parameter(0, *parameters)
+            print('zizizi',shift)
```
pmelchior, line 296: tss, tss, tss

scarlet/source.py
```diff
@@ -470,13 +470,11 @@ def __init__(
...
-                    zero=1e-20
-                )  # slightly positive values
-                step = partial(relative_step, factor=1e-6)
-                spectrum = Parameter(
-                    spectrum, name="spectrum", step=step, constraint=constraint,
-                )
+                noise_rms = np.concatenate(
+                    [np.array(np.mean(obs.noise_rms, axis=(1, 2))) for obs in observations]
+                ).reshape(-1)
+                spectrum = TabulatedSpectrum(model_frame, spectrum, min_step=noise_rms)
```
pmelchior, line 476: OK, good catch, this is all in the default `TabulatedSpectrum`

scarlet/blend.py
herjy: I think I used that for some tests. Removed

scarlet/renderer.py
herjy: This never happened.

scarlet/renderer.py
herjy: Done, Let's wait to see if tests pass and we can merge. Warning though, you're going to hate how I unfied the classes. I removed the original one and made the `ShiftConvolutionRenderer` the new `ConvolutionRenderer`. It's git-dirty.

herjy: Tests fail, wait till I fix it.

scarlet/renderer.py
```diff
@@ -185,7 +185,6 @@ def __init__(self, data_frame, model_frame, convolution_type="fft", padding=10):
         ll = np.round(pixel_in_model_frame.min(axis=0)).astype("int")
         ur = np.round(pixel_in_model_frame.max(axis=0)).astype("int") + 1
         bounds = (ll[0], ur[0]), (ll[1], ur[1])
-        data_box = model_frame.bbox[0] @ Box.from_bounds(*bounds)
```
pmelchior, line 188: why is this removed? It used to set `self.slices`, which is apparently already gone, but it was needed below...

scarlet/renderer.py
```diff
@@ -194,18 +193,6 @@ def __init__(self, data_frame, model_frame, convolution_type="fft", padding=10):
...
         self.diff_kernel = fft.match_psf(psf_fft, model_psf_fft, padding=padding)
 
-    def get_model(self, *parameters):
-        def transform(model):
-            # restrict to observed channels
-            model_ = self.map_channels(model)
-            # convolve observed channels
-            model_ = self.convolve(model_)
-            # adjust spatial shapes
-            model_ = match_shape(model_, self.data_frame, self.slices)
```
pmelchior, line 204: This line preserved the output shape, i.e. if the data frame is larger/smaller than the model frame. Why is this not needed any longer?

pmelchior (approved):

PR #249, opened 2021-07-20
Little things

pmelchior:


scarlet/model.py
```diff
@@ -66,14 +68,34 @@ def __next__(self):
...
-            return parameters[i]
+        # during optimization: parameters are passed by autograd
+        if parameters:
+            parameters_ = parameters
+        else:
+            parameters_ = self.parameters
 
-        # otherwise use self
-        if i < len(self.parameters):
-            return self.parameters[i]
```
herjy, line 76: Was this actually a bug? `parameters` was only used when it was longer than `self.parameters`?

herjy (approved):
Nice, parameters are now searchable by names. I suggest we merge this and I can use it in my PR.

herjy:
Please punt on merging, there might be a bug.

scarlet/model.py
pmelchior: No. `parameters` is only an argument to the function during optimization, otherwise it accesses them through `self`. They are identical arrays, but the former has an `autograd` box around it.

fred3m (approved):
Looks good. Thanks for the doc updates and allowing the user to access parameters by name.

scarlet/model.py
herjy: Glad I asked, it seems like it poses a problem in the new version (I'm pulling the psf_shift, but it says `parameters` is an `ArrayBox`)
```~/Desktop/git_repos/scarlet/scarlet/model.py in get_parameter(self, i, *parameters)
     95             for p in parameters_:
     96                 print(p)
---> 97                 print(p.name)
     98             match = tuple(p for p in parameters_ if p.name == i)
     99             if len(match) == 1:

AttributeError: 'ArrayBox' object has no attribute 'name'```

pmelchior: good catch! Let me fix that

PR #250, opened 2021-07-26
bug fix_data_box

herjy:
This fixes a bug in the misalignment of multi-resolution images. 
We actually need to know the bounds of the model frame to build the slices of the renderer.


pmelchior:
I see. Can you please remove lines 18-20 from `renderer.py` to prevent this mistake from being repeated. `slices` should be a part of the derived renderer, not the base class.

When you push the changes, we should get travis to run the test (not sure why it was stuck earlier).

herjy:
Yeah, I wasn't sure about removing it from the base class, wondering if it could ever be useful (or rather not useless).

pmelchior (approved):

PR #262, opened 2022-02-02
Update version.py

herjy:
This bug comes back again and again. It just requires keeping git up to date. I just made the error message more explicit.


pmelchior:
Thanks!

PR #264, opened 2022-02-09
Starlet mono

pmelchior:
Implements monotonicity constraints for starlet sources.

Currently, it's applied to _every_ starlet scale, which may be a bad idea, but this can be addressed later.


scarlet/morphology.py
```diff
@@ -283,6 +299,39 @@ def get_model(self, *parameters):
...
         return starlet_reconstruction(coeffs)
 
+    def update(self):
+        coeffs = self.get_parameter(0)
+        if coeffs.fixed:
+            return
+
+        # shrink the box?
+        image = self.get_model()
+        # image = proxmin.operators.prox_soft(image, 0, thresh=1e-6, type="absolute")
```
herjy, line 309: remove

scarlet/morphology.py
```diff
@@ -46,6 +48,24 @@ def __init__(self, frame, *parameters, bbox=None):
 
         super().__init__(*parameters)
 
+    def shrink_box(self, image, thresh=0):
```
herjy, line 51: Feels like this should belong in the Box class, just after the `grow` method.

scarlet/morphology.py
```diff
@@ -283,6 +299,39 @@ def get_model(self, *parameters):
...
+
+        # shrink the box?
+        image = self.get_model()
+        # image = proxmin.operators.prox_soft(image, 0, thresh=1e-6, type="absolute")
+        bbox = self.bbox.copy()
+        self.shrink_box(image, thresh=1e-8)
+        if bbox != self.bbox:
+            slice, _ = overlapped_slices(bbox, self.bbox)
+            center = tuple(s // 2 for s in self.bbox.shape)
+            constraint = MonotonicMaskConstraint(center, center_radius=1)
```
herjy, line 315: This seems to only work for monotonic constraints. shouldn't it cover the case where constraints are positivity andd L0? Havinng constraint being a class attribute would make it easy.

herjy (commented):
Nothing big just a few comments.

scarlet/morphology.py
pmelchior: not quite. `shrink_box` depends on the image (to find out by how much it can shrink), whereas `box.grow` just takes a radius.

That brings up a concern: `box.grow` does not need to respect the fixed set of box sizes we allow to keep the cache small. @fred3m have you checked, or is that not a concern?

scarlet/morphology.py
pmelchior: good catch!

I'm not sure what you mean to have the constraint a class attribute. in this case: which class should hold it?

Or put differently, this problem is the reason why I dislike too many options in source classes. I now have to store whether this class actually is `monotonic`.

scarlet/morphology.py
fred3m: This is a different method than growing the `Box`. `Box.grow` takes a `radius` and the method is used to increase the size of the box by that amount in each direction (top, bottom, left, right) but keep the same center. It is not automatically fit, so it's up to the user to restrict the size if they wish to do so (in my case this is not an issue because I only grow the box to create a new image that contains the source convolved with the PSF).

Since this is fitting the box size, and the algorithm to do that will be dependent on the use case, I think that it's more appropriate to keep it outside of the Box` class.

scarlet/morphology.py
```diff
@@ -110,51 +130,36 @@ def get_model(self, *parameters):
...
-            # adjust bbox
-            self.bbox.origin = tuple(o + dist for o in self.bbox.origin)
-            self.bbox.shape = (newsize, newsize)
             raise UpdateException
 
         # grow the box?
         # because the PSF moves power across the box, the gradients at the edge
         # accummulate flux from beyond the box
-        if image.m is not None:
+        elif image.m is not None:
```
fred3m, line 162: No need to use an `elif`, since the previous `if` will always raise an exception.

scarlet/morphology.py
herjy: I hesitated making the distinction the first time: I don't suggest to put the `shrink_box` method in `Box`, but to have a `shrink` method in `Box` that we would call here, and that we could call anytime we need boxes to shrink.

scarlet/morphology.py
herjy: Sorry, I meant making constraint an attribute of the class so that in the update you have `constraint=self.constraint` on line 323 (now)

scarlet/morphology.py
herjy: Though it can be done in a separate PR. No need to hold this one here, but I feel like it corresponds more to the spirit of how boxes work and also could possibly be useful for other morphologies.

scarlet/morphology.py
```diff
@@ -110,51 +130,36 @@ def get_model(self, *parameters):
...
+                v=image.v[slice] if image.v is not None else None,
+                vhat=image.vhat[slice] if image.vhat is not None else None,
             )
 
             # set new parameters
             self._parameters = (image,) + self._parameters[1:]
 
-            # adjust bbox
-            self.bbox.origin = tuple(o + dist for o in self.bbox.origin)
-            self.bbox.shape = (newsize, newsize)
```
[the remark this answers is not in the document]
pmelchior: that's now handled in `Morphology.shrink_box`. The reason for moving it there is that it can be used for `Starlet` and `Image` morphologies.

scarlet/morphology.py
pmelchior: that's why it's implemented at the `Morphology` level, not in `ImageMorphology`

scarlet/morphology.py
pmelchior: nitpicking, are we?

PR #265, opened 2022-02-11
Fix string in version.py

aboucaud:


version.py
```diff
@@ -24,7 +24,7 @@ def get_public_version():
         return _version
     except FileNotFoundError:
         msg = ("Could not find either a git repo or pre-installed version." +
-               "This should never happen unless your version of git is out of date. Please update git." + "
```
herjy, line 27: Hillarious! I might be the culprit since I tweaked these lines recently.

herjy (approved):

version.py
aboucaud: Didn't want to put you on the spot :)

Issue #222, opened 2020-11-05
Ringing in low magnitude galsim sims


herjy:
I recently noticed a significant ringing in the residuals of fits to very bright galaxies simulated with galsim. 
I intially thought that mono was not the culprit, but it seems that the proxes are making things a lot worse. 
Switching ALL constraints off makes things a lot better, but I'm not convinced that there isn't something wrong. Then each constraint, except for `CenterOnConstraint` does somnething awful.
No constraint & `CenterOnConstraint`:
<img width="1002" alt="Screen Shot 2020-11-05 at 6 04 23 PM" src="https://user-images.githubusercontent.com/16084926/98306206-5e6f3f00-1f91-11eb-8cd0-e8bffbf63566.png">
`Positivity` only:
<img width="1000" alt="Screen Shot 2020-11-05 at 6 05 17 PM" src="https://user-images.githubusercontent.com/16084926/98306283-83fc4880-1f91-11eb-8c9a-eb95968d6813.png">
`Normalization("max")` only:
<img width="995" alt="Screen Shot 2020-11-05 at 6 06 54 PM" src="https://user-images.githubusercontent.com/16084926/98306390-c160d600-1f91-11eb-863a-deabb393a819.png">
`monotonic` only:
<img width="998" alt="Screen Shot 2020-11-05 at 6 08 06 PM" src="https://user-images.githubusercontent.com/16084926/98306471-ed7c5700-1f91-11eb-81c8-37004ac1df44.png">

herjy:
It hints again at an issue with band-limitedness. In particular, i looks like the edges create some issues.

herjy:
Note, I get the same results with gaussians generated fron `scarlet.GaussianPSF`

pmelchior:
Good example. This is the same problem as in #215. As we have said there, it is unlikely that errors in the diff kernel itself have enough power to do that. With this new test, it's indeed more likely that undersampling is responsible.

Here's what I see. Both positivity and monotonicity become important in the outskirts and can change pixel values there even if the gradients are bandlimited (which they are). Thus, their artifacts look similar. But it seems that both have a step within the significant region of the source. I have no idea why because the proxes don't do anythere there.

[...]

PR #268, opened 2022-05-02
Leaky prox

pmelchior:
implements a class that wraps around a constraint to make the prox leaky. This is to prevent the optimization to get stuck when the subgradients exactly cancel the gradients and can be beneficial especially early on in the optimization.

This is an experimental feature that is not used in any standard source type, but may help with issues like #222


pmelchior:
this is a clean-up PR. Without travis to run tests and no side effects to existing code, I will merge without review

PR #286, opened 2024-07-26
Change get_affine to recognize wcs.wcs.cd

charlotteaward:
Many cameras (ZTF, DECam) need their wcs to be extracted as wcs.wcs.cd in the current scarlet setup. I've added a catch for this in interpolation.py.


pmelchior (approved):
please check if this fix is needed for scarlet2 as well

pmelchior:
You're free to squash&merge

